PR #20, opened 2019-03-07
Use hugo-theme-learn, plus lots of style and organization changes

yusefnapora:
This replaces the custom theme that was carried over from docs.ipfs.io with [hugo-theme-learn](https://learn.netlify.com/en/) a documentation focused theme that I think is pretty sweet.

As part of making the theme work, I also ended up restructuring things a bit, so hopefully the menu structure works a bit better. There are a lot of stub articles, but they should all at least link to an issue in the repo so we can track and prioritize the work.

screen shot 2019-03-08 at 11 41 30 am

I also got a bit carried away and put in a concept article on transports here, and a "getting started in js" thing I wrote up last week. Those would probably be better in separate PRs, but the site feels a little less empty with them, so I might just leave them in.  Going forward I'm going to aim for a PR per article to make things easier to review and correlate with issues.

Also, I got CI working! Thanks to help from @scout and @olizilla this thing builds on circle now and publishes a build of the site on each PR commit.

To see the preview, expand the "Show all checks" link at the bottom of the PR page and click Details on the IPFS one:

screen shot 2019-03-08 at 11 46 48 am


olizilla:
I love it! 

My only request from a "reading lots of text" perspective is, how about using the [system ui font](https://css-tricks.com/snippets/css/system-font-stack/#article-header-id-1) rather than Work Sans, for the body text? Work looks nice, but I'd rather read long docs in a more vanilla font.

yusefnapora:
Man, that's a good call @olizilla - it feels so much easier to read with the system text. I kept the custom font for the headings, since it looks nice and kind of breaks up the visual flow a bit.

yusefnapora:
I'm gonna roll forward with this thing - we can always fix it up in post 😄

Issue #4, opened 2019-02-20
Relays and AutoRelay


ghost:
We should address this topic somewhere in the docs, probably as part of a larger discussion of NATed hosts.  Points to cover:

1.  What are Relay nodes?  (Servers operated by PL that NAT'd hosts can connect to directly to discover peers.  Explain that *all* traffic from nodes using a relay will pass through the relay, which is expensive and expected to be slower than direct connections.  We should also explain to folks that if you use this, you're relying on a piece of centralized infrastructure and all the limitations that come with that...)

2.  How do I use relay nodes in my code?  How can a user tell whether his traffic is going through a relay node?  (I believe the answer to the first one is that it's just a flag in the go-libp2p Host constructor.  We should show a code example, including how to catch errors if relay nodes are unavailable for some reason.  Also, I believe only go-libp2p supports this right now; should make that clear.  On the second one, I don't know.  If there's not a way, we should log an issue with whoever is working on diagnostic tooling for libp2p.)

3.  Other NAT hole punching.  The whole reason for relay nodes is to help developers whose users have NAT'd hosts, like PCs on a home router.  We should roll this in to a larger discussion of NAT'd hosts, e.g., current status of uPnP to open ports on the router, current status of STUN/ICE (not available afaik), etc.

elahn:
I found the Circuit Relay > [Relay Addresses](https://docs.libp2p.io/concepts/circuit-relay/#relay-addresses) section confusing. Should some instances of `QmAlice` actually be `QmBob`?

jacobheun:
> I found the Circuit Relay > Relay Addresses section confusing. Should some instances of QmAlice actually be QmBob?

It's correct, but I think things could be clearer. There's really no reason to even mention `QmBob` because they're the dialing node and we don't really care what their ID is and I think it just adds to the confusion. 

Perhaps it would be better to simply take the recommended address structure `/ip4/7.7.7.7/tcp/55555/p2p/QmRelay/p2p-circuit/p2p/QmAlice` and break down the address components.

p-shahi:
close this in favor of: https://github.com/libp2p/docs/issues/172 and https://github.com/libp2p/docs/issues/171

PR #24, opened 2019-03-12
Concept docs: nat traversal and relay

### LINKED ISSUE #14: Write concept doc about NAT traversal and AutoNAT
NAT traversal is a pain point for every real-world p2p system, so we should highlight how libp2p addresses it "out of the box".

Should start with overview of the problem (what is NAT & why it sucks for p2p) and cover:

- automatic port forwarding via UPnP
- overview of STUN & description of identify protocol 
  - link to [spec](https://github.com/libp2p/specs/pull/97) when merged
- overview of TURN & description of libp2p relay protocol
  - link to [spec](https://github.com/libp2p/specs/tree/master/relay)

AutoRelay might deserve its own article, but could fit in here if we can link to https://github.com/libp2p/docs/issues/13 for details on how the DHT fits in.

yusefnapora:
This gives an overview of NAT and the ways libp2p tries to manage NAT traversal.

@Stebalien I based most of the NAT doc on this comment: https://github.com/libp2p/go-libp2p/issues/375#issuecomment-407122416 - if you could give this a look to see if I got things right that would be sweet.

I started to put the Circuit Relay explanation in the NAT doc, but it seemed to work better on its own, since relays are useful for more than NAT traversal.

This stuff is fairly light on details about how to enable or actually use relays, which is intentional, since that's quite implementation-specific. I'd like to keep the concept docs out of implementation land where possible.  That said, when we do have more usage docs, it would be good to have links back and forth from here to there.

Closes #14 and partially addresses #4 (with the exception of usage examples).

### DIFF
--- a/content/concepts/circuit-relay.md
+++ b/content/concepts/circuit-relay.md
@@ -3,34 +3,35 @@ title: Circuit Relay
 weight: 3
 ---
 
+Circuit relay is a [transport protocol](/concepts/transport/) that routes traffic between two peers over a third-party "relay" peer.
 
-In some cases, peers will be unable to [traverse their NAT](/concepts/nat/) in a way that makes them publicly accessible. Or they may not share common [transport protocols](/concepts/transport/) that would allow them to communicate directly.
+In many cases, peers will be unable to [traverse their NAT](/concepts/nat/) in a way that makes them publicly accessible. Or they may not share common [transport protocols](/concepts/transport/) that would allow them to communicate directly.
 
-To enable peer-to-peer architectures in the face of connectivity barriers like NAT, libp2p [defines a protocol called p2p-circuit][spec_relay]. This "circuit relay" protocol uses an intermediary peer to shuffle traffic between two peers that are unable to communicate directly.
+To enable peer-to-peer architectures in the face of connectivity barriers like NAT, libp2p [defines a protocol called p2p-circuit][spec_relay]. When a peer isn't able to listen on a public address, it can dial out to a relay peer, which will keep a long-lived connection open. Other peers will be able to dial through the relay peer using a `p2p-circuit` address, which will forward traffic to its destination.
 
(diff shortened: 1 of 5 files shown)


content/concepts/circuit-relay.md
```diff
@@ -0,0 +1,58 @@
+---
+title: Circuit Relay
+weight: 3
+---
+
+
+In some cases, peers will be unable to [traverse their NAT](/concepts/nat/) in a way that makes them publicly accessible. Or they may not share common [transport protocols](/concepts/transport/) that would allow them to communicate directly.
```
vyzo: Not just "some" cases, it's most of the time unfortunately.

content/concepts/circuit-relay.md
```diff
@@ -0,0 +1,58 @@
...
+
+To enable peer-to-peer architectures in the face of connectivity barriers like NAT, libp2p [defines a protocol called p2p-circuit][spec_relay]. This "circuit relay" protocol uses an intermediary peer to shuffle traffic between two peers that are unable to communicate directly.
```
vyzo: perhaps add a reference to TURN here.

content/concepts/circuit-relay.md
```diff
@@ -0,0 +1,58 @@
...
+
+`/ip4/7.7.7.7/tcp/55555/p2p/QmRelay/p2p-circuit/p2p/QmAlice`
+
+Everything prior to the `/p2p-circuit/` above is the address of the relay peer, which includes the transport address and their peer id `QmRelay`. After `/p2p-circuit/` is the peer id for my peer at the other end of the line, `QmAlice`.
+
+By giving the full relay path to my friend `QmBob`, they're able to quickly establish a relayed connection without "ask around" for a relay that has a route to `QmAlice`.
+
+#### AutoRelay
+
+The circuit relay protocol is only effective if peers can discover willing relay peers that are accessible to both sides of the relayed connection.
```
vyzo: and also advertise the relay addresses.

PR #42, opened 2019-04-17
Concept doc for "libp2p protocols"

yusefnapora:
This adds a concept doc for what I'm calling "libp2p protocols", meaning protocols built with libp2p and routed by the switch/swarm, etc.

This covers the basic features of a protocol, talks a bit about protocol negotiation without getting too deep into multistream-select, and goes over the "built-in" libp2p protocols.

Some questions:

- is there a better name than "libp2p protocol" I can use to distinguish these from protocols not built with libp2p?
- are there any core libp2p protocols I'm missing?
  - apart from pubsub... I want to stub out the pubsub article first and link to it
- is it misleading to call these "core" or "built-in" protocols? A case could be made for ping and identify, but you can definitely use libp2p without the DHT, for example. And some things like circuit relay are completely optional, but are still part of the "libp2p family" of protocols.


content/concepts/protocols.md
```diff
@@ -0,0 +1,268 @@
...
+## Core libp2p protocols
+
+In addition to the protocols that you write when developing a libp2p application, libp2p itself defines several foundational protocols that are used for core features.
+
+### Common patterns
+
+The protocols described below all use [protocol buffers](https://developers.google.com/protocol-buffers/) (aka protobuf) to define message schemas.
+
+Messages are exchanged over the wire using a very simple convention called [msgio](https://github.com/jbenet/go-msgio), which prefixes binary
+message payloads with an integer that represents the length of the payload in bytes. The length is encoded as a [protobuf varint](https://developers.google.com/protocol-buffers/docs/encoding#varints)  (variable-length integer).
```
vyzo: We are not actually using msgio afaict, but we do use varint lengh-prefixed messages.

content/concepts/protocols.md
```diff
@@ -0,0 +1,268 @@
...
+| `/ipfs/id/1.0.0` | [identify spec][spec_identify] | [go][identify_go] | [js][identify_js] | [rust][identify_rust] |
+
+<!-- TODO(yusef): update spec link on PR merge -->
+[spec_identify]: https://github.com/libp2p/specs/pull/97/files
+[identify_go]: https://github.com/libp2p/go-libp2p/tree/master/p2p/protocol/identify
+[identify_js]: https://github.com/libp2p/js-libp2p-identify
+[identify_rust]: https://github.com/libp2p/rust-libp2p/tree/master/protocols/identify/src
+
+The `identify` protocol allows peers to exchange information about each other, and serves a crucial role
+in [NAT traversal](/concepts/nat/).
```
vyzo: that's probably overstating `identify`'s role in nat traversal, which is minimal.

PR #63, opened 2019-08-27
fix: js ping tutorial

vasco-santos:
With the latest refactor released with [libp2p/js-libp2p400](https://github.com/libp2p/js-libp2p/pull/400), [libp2p/js-libp2p-ping](https://github.com/libp2p/js-libp2p-ping) module is now deprecated.


vasco-santos:
@yusefnapora do you happen to know if there is something wrong with CI? 
I don't have permissions to restart it, so I am not sure if this was a problem or not

yusefnapora:
@vasco-santos looks like it was a transient issue pinning to the cluster, but it's hard to say what the cause was...

PR #85, opened 2020-09-12
Ping responders don't close the stream

tmakarios:
Ironically, I think rust-libp2p used to close the stream after responding to a ping, until [they concluded](https://github.com/libp2p/rust-libp2p/issues/1601) that this behaviour didn't conform to [the non-existent specification](https://github.com/libp2p/specs/issues/183).


jacobheun (approved):
lgtm, they're indeed not closed.

Issue #80, opened 2020-05-05
Broken link Configure libp2p "one for node.js" "one for the browser"


haydenyoung:
Thanks for the great documentation.

https://docs.libp2p.io/tutorials/getting-started/javascript/#configure-libp2p

The links "one for node.js" and "one for the browser" do not link to an existing Github page.

https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/libp2p-nodejs.js

https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/libp2p-browser.js

Guessing their new location is:

https://github.com/ipfs/js-ipfs/blob/10e8682e44ed06fdcad849b7cb7e6a38dbd9ba49/packages/ipfs/src/core/runtime/libp2p-nodejs.js

https://github.com/ipfs/js-ipfs/blob/10e8682e44ed06fdcad849b7cb7e6a38dbd9ba49/packages/ipfs/src/core/runtime/libp2p-browser.js

jacobheun:
Yes those are the correct new links. js-ipfs recently consolidated several of its packages, which looks to have broke these. Would you mind submitting a PR with the new links?

We are planning to improve configuration and have defaults available directly in js-libp2p in the future, https://github.com/libp2p/js-libp2p/issues/576, which will help us mitigate the broken references.

PR #86, opened 2020-10-02
Fixed broken links on js-libp2p getting started tutorial 

EthanTheMaster:
This PR addresses Issue #80. It seems that the links to the `libp2p-nodejs.js` and `libp2p-browser.js` files constantly change, but hopefully the links that I put will not break in the future. Additionally, there is a broken link in the page that points to the definition of a multiaddress. As a fix, I instead pointed the link to the glossary entry for the multiaddress. Finally the output in the terminal shows the `^C` control character which I am guessing came from copying the terminal output after killing the ping server. This control character was removed from the output.

### DIFF
--- a/content/tutorials/getting-started/javascript.md
+++ b/content/tutorials/getting-started/javascript.md
@@ -44,7 +44,7 @@ Side note: throughout this tutorial, we use the `> ` character to indicate your 
 
 libp2p is a very modular framework, which allows javascript devs to target different runtime environments and opt-in to various features by including a custom selection of modules.
 
-Because every application is different, we recommend configuring your libp2p node with just the modules you need. You can even make more than one configuration, if you want to target multiple javascript runtimes with different features. For example, the IPFS project has two libp2p configurations, [one for node.js](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/libp2p-nodejs.js) and [one for the browser](https://github.com/ipfs/js-ipfs/tree/master/src/core/runtime/libp2p-browser.js).
+Because every application is different, we recommend configuring your libp2p node with just the modules you need. You can even make more than one configuration, if you want to target multiple javascript runtimes with different features. For example, the IPFS project has two libp2p configurations, [one for node.js](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core/src/runtime/libp2p-nodejs.js) and [one for the browser](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core/src/runtime/libp2p-browser.js).
 
 {{% notice note %}}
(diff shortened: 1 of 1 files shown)


content/tutorials/getting-started/javascript.md
```diff
@@ -139,7 +139,7 @@ const node = await Libp2p.create({
 
 Now that you have configured a **Transport**, **Crypto** and **Stream Multiplexer** module, you can start your libp2p node. We can start and stop libp2p using the [`libp2p.start()`](https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#start) and [`libp2p.stop()`](https://github.com/libp2p/js-libp2p/blob/master/doc/API.md#stop) methods.
 
-A libp2p node needs to have a listen address for the given transport, so that it can be reached by other nodes in the network. Accordingly, we will install the `multiaddr` module to create a tcp [multiaddress](definition_muiltiaddress) and add it to the node.
+A libp2p node needs to have a listen address for the given transport, so that it can be reached by other nodes in the network. Accordingly, we will install the `multiaddr` module to create a tcp [multiaddress](https://docs.libp2p.io/reference/glossary/#multiaddr) and add it to the node.
```
vasco-santos, line 142: I believe the issue is below in this file. The definition of `definition_multiaddress` is linking into: https://docs.libp2p.io/tutorials/getting-started/javascript/definition_muiltiaddress but it should be starting from the origin path https://docs.libp2p.io/reference/glossary/#multiaddr

EthanTheMaster: Sorry, I don't think I'm following what you're suggesting. What do you mean by "it should be starting from the origin path ..."? Was the link that I used pointing into the glossary entry for the multiaddress not the intended link? Also, I guess I hadn't notice this before when I changed the link but it seems that `definition_muiltiaddress` was misspelled (there is an `i` between the `u` and `l`), but even after correcting the spelling, the link is still broken.

vasco-santos: I mean that there is a reference for it in the end of the document: https://github.com/libp2p/docs/pull/86/files#diff-2a666e46d0d9d65d1a1c33e40635e1b2R280

We should change it there. Ideally not using the entire url but a local path

EthanTheMaster: Ok, I see what you mean! Yeah the link to the glossary page should probably be a relative path instead of an absolute path to the page for stability. It seems that using references in markdown requires the format `[text][ref]` as opposed to `[text](ref)` which was causing the error. Hopefully, the fix I just did resolves that broken link issue.

vasco-santos (approved):
LGTM! Thanks for the fixes

PR #88, opened 2020-12-11
chore: update js tutorial switching secio to noise

vasco-santos:
We deprecated secio, but it is still in the docs website


jacobheun (approved):
lgtm

PR #94, opened 2021-02-25
Fixed IRC link

ardunster:
The IRC channel instructions are actually at IPFS, updated the link so it doesn't try to reference a non-existing page in libp2p docs.


DannyS03:
replaced by: https://github.com/libp2p/docs/pull/163

PR #97, opened 2021-05-18
Added the missing code example instructing to npm install the multiad…

simon-o-matic:
…dr module

The text says you have to do this, but doesn't show you how as with the other modules. So the example doesn't run if you just copy all the code. Now it does.


yusefnapora:
thank you!

Issue #55, opened 2019-06-25
Broken link to PeerId in Transport Concepts docs


tesla809:
Inside the [Transport docs](https://docs.libp2p.io/concepts/transport/) in Concepts section, the following sentence has a broken link:

“When dialing a remote peer, the multiaddress should include the [PeerId](https://docs.libp2p.io/concepts/peer-identity/) of the peer you’re trying to reach”

The PeerId link goes to: https://docs.libp2p.io/concepts/peer-identity/
Which returns the message:
`ipfs resolve -r /ipns/docs.libp2p.io/concepts/peer-identity/: no link named "peer-identity" under QmcuoK1NLYx6n438xBvxqH1nfnj7LTdSQxKcKCX2wM9x6n`

tesla809:
However, this link works:
https://docs.libp2p.io/concepts/peer-id/

jochasinga:
The [Peer Identity](https://docs.libp2p.io/concepts/addressing/peer-id) link on [this page](https://docs.libp2p.io/concepts/addressing/) is also broken.

mxinden:
Closing here since links have been fixed. Thanks for the help @tesla809 and @jochasinga!

PR #109, opened 2021-11-12
Fix broken Peer ID link on Addressing page

jochasinga:
The link previously led to https://docs.libp2p.io/concepts/addressing/peer-id/.
Now fixed so it leads to https://docs.libp2p.io/concepts/peer-id/.

See also: #55 

FYI: @momack2


momack2:
Thanks Pan!!

PR #116, opened 2021-12-21
For js-libp2p node >= v16 is required

TimDaub:
See https://github.com/libp2p/js-libp2p/issues/1049#issuecomment-987080135


vasco-santos (approved):
Thanks @TimDaub  ❤️

PR #124, opened 2022-01-06
fix(link): Go example links

sleeping-barber:
Hej;
I recognized that the current documentation is pointing to an archived repository. This PR redirects the user to the correct location without any hop in between.

Cheers!


marten-seemann (approved):
Thanks!

PR #127, opened 2022-01-29
mention that peer discovery can be accomplished with kademlia and identify

Frederik-Baetens:
This would've helped point me into the right direction a bit earlier. If there's better places to mention this, or if a more extensive explanation is wanted, I'd be happy to expand on this pull request.

### DIFF
--- a/content/concepts/publish-subscribe.md
+++ b/content/concepts/publish-subscribe.md
@@ -58,7 +58,7 @@ peers on its behalf, a process called **ambient peer discovery**.
 
 Potential methods for discovering peers include:
 
-* Distributed hash tables
+* Distributed hash tables (e.g. by combining libp2p's kademlia DHT with its identify protocol)
 * Local network broadcasts
 * Exchanging peer lists with existing peers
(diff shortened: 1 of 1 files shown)


mxinden:
First of all, thanks for helping out here!

---

In this particular case, I think this is rather a shortcoming of the rust-libp2p implementation than a shortcoming of the documentation. If I am not mistaken the other libp2p implementations see the identify protocol as a core protocol. Thus a user does not have to hook up the Kademlia protocol to the identify protocol.

In rust-libp2p we try to stay as generic as possible. E.g. in this case not assume the identify protocol always exist and thus requiring users to hook it up manually to the Kademlia protocol in case it does exist.

To avoid confusion for users of other implementations, I would suggest moving this addition to the rust-libp2p documentation, e.g. to the crate level documentation of `libp2p-identify`. This in combination with your example (https://github.com/libp2p/rust-libp2p/pull/2457) should be useful for newcomers.

Frederik-Baetens:
Thanks for the feedback, i'll open a PR to document this on the rust-libp2p side then!

PR #128, opened 2022-01-31
add an explanation about boot nodes to the glossarry

Frederik-Baetens:
I think this is useful to put in the glossary, so that I can refer to it from other places in the documentation.

### DIFF
--- a/content/reference/glossary.md
+++ b/content/reference/glossary.md
@@ -5,8 +5,9 @@ weight: 1
 
 ### Boot Node
 
-New nodes in a p2p network often make their initial connection to the p2p network through a set of nodes known as boot nodes. Information about these boot nodes is shared beforehand outside of the p2p network, for example by posting the IP addresses of the boot nodes on a website.
-After connecting to the boot nodes, the new node can reach out to other nodes in the network and connect directly to them, thereby no longer relying on the boot nodes.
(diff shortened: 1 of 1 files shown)


content/reference/glossary.md
```diff
@@ -3,6 +3,11 @@ title: "Glossary"
 weight: 1
 ---
 
+### Boot Node
+
+New nodes in a p2p network often make their initial connection to the p2p network through a set of nodes known as boot nodes. Information about these boot nodes is shared beforehand outside of the p2p network, for example by posting the IP addresses of the boot nodes on a website.
```
mxinden, line 8:
```suggestion
New nodes in a p2p network often make their initial connection to the p2p network through a set of nodes known as boot nodes. Information (e.g. addresses) about these boot nodes is e.g. embedded in an application binary or provided as a configuration option.
```

content/reference/glossary.md
```diff
@@ -3,6 +3,11 @@ title: "Glossary"
...
+After connecting to the boot nodes, the new node can reach out to other nodes in the network and connect directly to them, thereby no longer relying on the boot nodes.
```
mxinden, line 9:
```suggestion

The boot nodes serve as an entry point, providing a list of other nodes in the network to newcomers. After connecting to the boot nodes, the new node can connec to those other nodes in the network, thereby no longer relying on the boot nodes.
```

mxinden (commented):
Cool. Good idea. Two suggestions. Otherwise this looks good to me.

Frederik-Baetens:
:+1: implemented your suggestions.

mxinden (approved):
Thanks!

PR #129, opened 2022-02-16
Fix: remove dead link from Glossary

jchris:
Remove dead link from docs -- the repo it points to has been deprecated.


vasco-santos (approved):
LGTM
That is now part of js-libp2p core

Issue #100, opened 2021-06-22
secio is deprecated


marten-seemann:
`content/concepts/protocols.md` still mentions secio. We should update it with references to TLS 1.3 and Noise.

PR #130, opened 2022-02-16
Modify secio to say it is deprecated

momack2:


momack2:
fixes https://github.com/libp2p/docs/issues/100

PR #138, opened 2022-03-19
Update go.md to make tutorial compliable and runnable. 

tristanisham:
In the original tutorial there was no mention of the need to create a context. As such, the original tutorial had an undeclared `ctx` variable. 

I've added the minimum required context for the full program to compile. Without it, the tutorial "full program" will not build or run.


marten-seemann:
This PR is wrong. The libp2p constructor doesn't take a context (any more).

PR #142, opened 2022-03-25
added missing multiaddr require

danicuki:


mxinden:
@achingbrain or @vasco-santos mind taking a look?

vasco-santos:
@danicuki thanks for the PR. We don't need to install multiaddr and that is the outdated part of this docs. Could you update the PR reverting the proposed change and removing the information about needing to install the module?

danicuki:
In the `Lets play ping-pong` session, there is a explicit call to `multiaddr` function, so I don't understand well the claim. How this line would be executed without importing `multiaddr` function?

`const ma = multiaddr(process.argv[2])`

achingbrain:
@danicuki you are correct in that the `multiaddr` function is invoked, but it might make more sense to the reader to add the `require` to the block in the `Let's play ping pong!` section where the `process` module is `require`d e.g.:

> First, require the `process` module so that we can get the command line arguments *, and the multiaddr module to validate the address supplied by the user*:
>
>```javascript
>const process = require('process')
>const { multiaddr } = require('multiaddr')
>```
>
>Then we'll need to parse the multiaddress...

danicuki:
added a new PR to solve it:

https://github.com/libp2p/docs/pull/145

PR #143, opened 2022-03-28
Fix link for Protocol Negotiation

flaki:
The Transports concepts page links to the "Protocol Negotiation" page but this hase moved under the "Protocol" page under its own heading.


mxinden (approved):
Thanks!

PR #148, opened 2022-05-18
chore: update hugo to latest & add m1 mac support

yusefnapora:
This updates the hugo version downloaded by the `Makefile` to the latest `0.99.1`, from the ancient `0.54.0` that we were using. I also added a check for the `arm64` architecture, so it will download the right version if you're on an m1 mac (or linux on arm, I guess).

I had to pull down the latest version of the `hugo-theme-learn` we're using, and I also made a couple changes to the layout partials in this repo, since some of the variables they were using have been deprecated and removed.

@walkerlj0 asked if I could show her how to fix this, so I made [a screencast of me fiddling with things until they work](https://drive.google.com/file/d/1fulHguKRTpfxCBm36ePmGrwqbFU7RZQ4/view?usp=sharing) 😄

**Note** if you've already cloned this repo and run `make` in the past, you'll need to delete `bin/hugo` after updating to this PR.


mxinden (approved):
This looks good to me. (Can't validate on an M1, nor other ARM processor though.)

Thaks @yusefnapora!

walkerlj0:
@mxinden @yusefnapora pulled in this branch, can confirm it works on M1, approved!

PR #160, opened 2022-07-30
Add dos and monitoring docs

MarcoPolo:
Addresses https://github.com/libp2p/go-libp2p-resource-manager/issues/31.

Adds some documentation and links for libp2p monitoring and dos mitigation.


content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+weight: 4
+---
+
+In general we have to design our protocols to be resilient to malicious or
+misbehaving nodes. libp2p offers reasonable defaults and primitives tools let the
+end application developer design resilient peer-to-peer programs.
+
+# General tactics that work across libp2p implementations
+
+## Limit the number of concurrent streams your protocol needs
```
mxinden, line 12: Optional: I think a similar section on the number of connections would be helpful.

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+
+If you can split up your libp2p application into multiple separate processes you
+can increase the resiliency of your overall system. For example your node may
+have to help achieve consensus and respond to user queries. By splitting this up
+into two processes you now rely on the OS’s guarantee that the user query
+process won’t take down the consensus process.
+
+## Fail2ban
+
+If you can log when a peer is misbehaving or malicious, you can then hook up
```
mxinden, line 35:
```suggestion
If you can log when a peer is misbehaving or is malicious, you can then hook up
```

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+with a special format. For example here’s a peer status log line that tells us a
+peer established a connection with us, and that this log line was randomly
+sampled (1 out of 100).
+
+```
+Jul 27 12:14:14 ipfsNode ipfs[46133]: 2022-07-27T12:14:14.674Z        INFO        canonical-log        swarm/swarm_listen.go:128        CANONICAL_PEER_STATUS: peer=12D3KooWSbNLGMYeUuMSXDiHwbhXHzTJaWZzH95MZzeAob9BeB51 addr=/ip4/147.75.74.239/udp/4001/quic sample_rate=100 connection_status="established" dir="inbound"
+```
+
+To see these kinds of logs make sure you’ve enabled the `"canonical-log=info"`
+log level. You can this in code like
```
mxinden, line 85:
```suggestion
log level. You can do this in code like
```

content/reference/monitoring.md
```diff
@@ -0,0 +1,22 @@
+---
+title: "Monitoring and Observability"
+weight: 3
+---
+
+Monitoring your libp2p process lets make sure that in practice things are going
```
mxinden, line 6:
```suggestion
Monitoring your libp2p process makes sure that in practice things are going
```

mxinden (approved):
Thank you @MarcoPolo for writing this up!

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
+---
+title: "DOS Mitigation"
+weight: 4
+---
+
+In general we have to design our protocols to be resilient to malicious or
+misbehaving nodes. libp2p offers reasonable defaults and primitives tools let the
```
BigLep, line 7:
```suggestion
misbehaving nodes. libp2p offers reasonable defaults and primitives to let the
```

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+
+This section is meant for an operator of a go-libp2p application. In the future
+we hope that this advice will work equally well amongst all implementations.
+
+## Resource Manager
+
+go-libp2p includes a powerful resource manager that keeps track of every active
+connection and stream. It exposes some useful metrics that can help you tune
+your resource limits. It’s important to keep an eye on the number of blocked
+resource requests, if you are constantly blocking resources during normal
```
BigLep, line 51:
```suggestion
resource requests.  If you are constantly blocking resources during normal
```

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+operation your limits are too low. But a very rare spike is a good sign that the
+resource manager protected your node from something. You can hook these metrics
+up to Prometheus (or any other opencensus exporter) and Grafana (we even include
+a prebuilt dashboard). See
+[https://github.com/libp2p/go-libp2p-resource-manager/tree/master/obs/grafana-dashboards](https://github.com/libp2p/go-libp2p-resource-manager/tree/master/obs/grafana-dashboards)
```
BigLep, line 56:
```suggestion
https://github.com/libp2p/go-libp2p-resource-manager/tree/master/obs/grafana-dashboards
```

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+for more details on how to set this up.
+
+![blocked-resource-screenshot](./assets/blocked-resource-screenshot.png)
+
+## Rate limit incoming connections
+
+Depending on your use case, it can help to limit the number of inbound
+connections. You can use go-libp2p's
```
BigLep, line 64: Do we give guidance on when to use this mechanism vs. the resource manager?  I see this is a good hook for custom logic, but it seems like what Prysm is doing could be covered by go-libp2p resource manager right?

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+ip address of the peer from the
+[multiaddr](https://github.com/multiformats/multiaddr) in the log.
+
+```bash
+sudo ufw deny from 1.2.3.4
+```
+
+## How to automate blocking with fail2ban
+
+You can hook up [fail2ban](https://www.fail2ban.org/wiki/index.php/Main_Page) to
```
BigLep, line 102:
```suggestion
You can hook up [fail2ban](https://www.fail2ban.org) to
```

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+automatically block connections from these misbehaving peers if they emit this
+log line multiple times in some period of time. For example, a simple fail2ban
+filter for go-libp2p would look like this:
+
+```
+[Definition]
+failregex = ^.*[\t\s]CANONICAL_PEER_STATUS: .* addr=\/ip[46]\/<HOST>[^\s]*
```
BigLep, line 109: I had a question about this in https://www.notion.so/pl-strflt/Guide-for-how-to-respond-to-resource-exhaustion-attacks-b10f55cc9a3d4917ae80c9b914e05e8c.

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+
+This matches any canonical peer status logs. If a peer shows up often in these
+sampled logs, something abnormal is happening. i.e. maybe they are churning
+connections.
+
+A conservative fail2ban rule for go-libp2p using the above filter would look
+like this:
+
+```
+[go-libp2p-weird-behavior-iptables]
```
BigLep, line 121: I had some comments about this section and format/layout changes in I had a question about this in https://www.notion.so/pl-strflt/Guide-for-how-to-respond-to-resource-exhaustion-attacks-b10f55cc9a3d4917ae80c9b914e05e8c.

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+Environment=GOLOG_LOG_LEVEL="canonical-log=info" LIBP2P_RCMGR=1
+ExecStart=/nix/store/mmvd2akskpaszlradl8qv4v703v1cy11-kubo-0.0.1/bin/ipfs daemon
+Restart=always
+RestartSec=1min
+User=ipfs
+```
+
+### Example screen recording of fail2ban in action
+
+[fail2ban+go-libp2p.mov](./assets/fail2bango-libp2p.mov)
```
BigLep, line 165: Is .mov supported across platforms without quicktime?  Maybe do .mp4 instead?

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Setting Up fail2ban
```
BigLep, line 167: I made some edits in I had a question about this in https://www.notion.so/pl-strflt/Guide-for-how-to-respond-to-resource-exhaustion-attacks-b10f55cc9a3d4917ae80c9b914e05e8c. about doing this more as a numbered list.

content/reference/monitoring.md
```diff
@@ -0,0 +1,22 @@
...
+
+Look at the [libp2p-metrics crate](https://github.com/libp2p/rust-libp2p/tree/master/misc/metrics).
+
+Go libp2p:
+
+For resource usage take a look at the OpenCensus metrics exposed by the resource
+manager
+[here](https://pkg.go.dev/github.com/libp2p/go-libp2p-resource-manager@v0.5.2/obs).
+In general Go libp2p wants to add more metrics across the stack in the future,
+this work is being tracked in issue
```
BigLep, line 21:
```suggestion
In general Go libp2p wants to add more metrics across the stack. 
This is being tracked in 
```

content/reference/monitoring.md
```diff
@@ -0,0 +1,22 @@
...
+
+Rust libp2p:
+
+Look at the [libp2p-metrics crate](https://github.com/libp2p/rust-libp2p/tree/master/misc/metrics).
+
+Go libp2p:
+
+For resource usage take a look at the OpenCensus metrics exposed by the resource
+manager
+[here](https://pkg.go.dev/github.com/libp2p/go-libp2p-resource-manager@v0.5.2/obs).
```
BigLep, line 19: Is there a way to link to @latest somehow?

content/reference/dos-mitigation.md
BigLep: Agreed.

BigLep, line 12: To be clear that we're not talking about global number of streams for a protocol, but the number of streams per connection...

```suggestion
## Limit the number of concurrent streams per connection your protocol needs
```

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+Each stream has some resource cost associated with it. Depending on the
+transport and multiplexer, this can be bigger or smaller. Try to avoid having
+too many concurrent streams open per peer for your protocol. Instead try to
+limit the maximum number of concurrent streams to something reasonable (surely
+you don't need >512 streams open at once for a peer?). Multiple concurrent
+streams can be useful for logic or to avoid [Head-of-line
+blocking](https://en.wikipedia.org/wiki/Head-of-line_blocking).
+
+Using a stream for a short period of time and then closing it is fine. It's
+really the number of _concurrent_ streams that you need to be careful of.
```
BigLep, line 23: Maybe use Identify as an example, where it doesn't make sense for the protocol to support more than one stream per connection?

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Fail2ban
+
+If you can log when a peer is misbehaving or malicious, you can then hook up
+those logs to fail2ban and have fail2ban manage your firewall to automatically
+block misbehaving nodes. Go libp2p includes some builtin support for this
+usecase. More details below.
+
+
+# Tactics for Go libp2p
```
BigLep, line 41:
```suggestion
# Tactics for go-libp2p
```

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
+---
+title: "DOS Mitigation"
```
BigLep, line 2: Can we have a table of contents?  Or does the docs website autogenerate?  I didn't see one in https://bafybeiegvuphbhruoy6ntjhgh5xfmweye7ubeshinyv5lugdhgxml6dice.on.fleek.co/reference/dos-mitigation/

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+weight: 4
+---
+
+In general we have to design our protocols to be resilient to malicious or
+misbehaving nodes. libp2p offers reasonable defaults and primitives tools let the
+end application developer design resilient peer-to-peer programs.
+
+# General tactics that work across libp2p implementations
```
BigLep, line 10: Looks like we need to add a level of indentation throughout given the rendering in https://bafybeiegvuphbhruoy6ntjhgh5xfmweye7ubeshinyv5lugdhgxml6dice.on.fleek.co/reference/dos-mitigation/

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,197 @@
...
+block misbehaving nodes. Go libp2p includes some builtin support for this
+usecase. More details below.
+
+
+# Tactics for Go libp2p
+
+This section is meant for an operator of a go-libp2p application. In the future
+we hope that this advice will work equally well amongst all implementations.
+
+## Resource Manager
```
BigLep, line 46: This section seems incomplete.  At the minimum, we should be linking to the resource manager guide/docs.

content/reference/dos-mitigation.md
BigLep, line 56: It seems like more of this should be in monitoring.md and we link there.

content/reference/dos-mitigation.md
BigLep, line 167:
```suggestion
### Setting Up fail2ban
```

BigLep (commented):
Thanks for doing this @MarcoPolo .  It's great starting to get to see how this will be exposed to users.  

I'm copying in some of the comments I had in https://www.notion.so/pl-strflt/Guide-for-how-to-respond-to-resource-exhaustion-attacks-b10f55cc9a3d4917ae80c9b914e05e8c

The information above is really helpful/good.  Looking at this fresh, I think we need to paint a big picture.  I view these a layering that someone should think about.  I’m apt to order chronologically based on the lifecycle of a libp2p application from design, development, testing, and operating.

1. What are the steps I can take to protect my application?  Basically what are all the defenses I can have in place before my application is deployed?  
    1. (Should be thought about early) Architect application well with limiting blast radius as described above
    2. Give clear signal of problems as highlighted above by using canonicallog
    3. Setup/tune resource management for each libp2p node.  Here we can link to resource manager docs.  We have to make sure this is user-friendly.  There are followups in https://github.com/libp2p/go-libp2p-resource-manager/issues/68 to handle.
2. How do I know the health of my node?  This is the monitoring doc.
3. What are the steps I can take when being attacked?  If the steps above were taken, an attack shouldn’t fully compromise my node but it may degrade performance.  Followup actions may be needed.
    1. Ban offending IPs.  It’s best to automate this as shown above using fail2ban.
    2. Maybe something about using the denylist?

content/reference/dos-mitigation.md
MarcoPolo: > Do we give guidance on when to use this mechanism vs. the resource manager?

No, but I could add something here.

> I see this is a good hook for custom logic, but it seems like what Prysm is doing could be covered by go-libp2p resource manager right?

Not really. If you're trying to avoid an adversary that can connect to you and give you a ton of work to do all at once the rcmgr doesn't protect at all. This attack can easily be mitigated by rate limiting though.

Not all applications will want this rate limiting, or they may want to rate limit certain things (e.g. something in the protocol rather than in the connections). For example, if I'm Google I wouldn't want to rate limit any new connection to me. I would rather rate limit work per connection.

Should the rcmgr do this? I don't think so. It's not directly related to limiting the resources being used, and if it can be handled by a smaller component that already exists the better.

I hope that makes sense, but happy to expand more as well.

content/reference/monitoring.md
```diff
@@ -0,0 +1,7 @@
+---
+title: "Monitoring and Observability"
+weight: 4
+---
+
+Reference the [Monitoring your application](todo) section in [DOS
```
MarcoPolo, line 6: Link this

content/reference/dos-mitigation.md
MarcoPolo: added one

content/reference/dos-mitigation.md
MarcoPolo: Answered. Copied here:

> `<HOST>`: I assume this is fail2ban syntax for saying that the host to block is this string that comes after “/ip[46]/” and before “[^\s]”.  Is that right?  (I was a little surprised not to see more conventional capture group conventions.)

Exactly. fail2ban expands this to a regex that captures the host `(?:::f{4,6}:)?(?P<host>\S+)`. See https://www.fail2ban.org/wiki/index.php/MANUAL_0_8#Filters

content/reference/dos-mitigation.md
MarcoPolo: Answered 

> @Marco Munizaga : it’s not clear to me how the fail2ban rule below is tied to the filter above.  I don’t see an id or name reference.

By the filename. The filter is in a file go-libp2p-peer-status.conf and this rule references filter=go-libp2p-peer-status.

I'll leave a comment on the relevant line as well

content/reference/monitoring.md
MarcoPolo: Generally a preference for a link that will always work, but not a strong opinion here. I'll change it to point to latest.

content/reference/dos-mitigation.md
MarcoPolo: Thanks! I copied those changes over here

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,311 @@
...
+really the number of _concurrent_ streams that you need to be careful of.
+
+## Limit the number of connections your application needs
+
+Like streams, each connection has a resource cost associated with it. A
+connection will usually represent a peer and a set of protocols with each their
+own resource usage. So limiting connections can have a leveraged effect on your
+resource usage. 
+
+In go-libp2p the number of active connections is managed by the
```
BigLep, line 95: From reading through this doc I don't think it's clear for a user on when to use the connmgr or the resource manager for go-libp2p.

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,311 @@
...
+[`connmgr`](https://pkg.go.dev/github.com/libp2p/go-libp2p@v0.21.0/p2p/net/connmgr#BasicConnMgr.Protect).
+`ConnManager` will trim connections when you hit the high watermark number of
+connections. You can protect certain connections with the `.Protect` method.
+
+In rust-libp2p handlers should implement
```
BigLep, line 100: For Rust, I think we should draw more from https://www.notion.so/pl-strflt/rust-libp2p-2a62a76b60c54bd69aea2aa3760d6efe .  At the minimum we should get a link to https://docs.rs/libp2p/latest/libp2p/swarm/struct.ConnectionLimits.html .

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,311 @@
...
+limit the maximum number of concurrent streams to something reasonable (surely
+you don't need >512 streams open at once for a peer?). Multiple concurrent
+streams can be useful for logic or to avoid [Head-of-line
+blocking](https://en.wikipedia.org/wiki/Head-of-line_blocking), but having too
+many streams will offset these benefits.
+
+Using a stream for a short period of time and then closing it is fine. It's
+really the number of _concurrent_ streams that you need to be careful of.
+
+## Limit the number of connections your application needs
```
BigLep, line 88: Given connections happen before streams in an application's lifecycle, maybe move this above the section above?

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,311 @@
...
+The general strategy is to use the minimum amount of resources as possible, and
+make sure that there's no untrusted amplification mechanism (e.g. an untrusted
+node can force you do to 10x the work it does). A protocol level reputation
+system can help (take a look at [GossipSub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) for inspiration) as well as
+logging misbehaving nodes and actioning those logs separately (see fail2ban
+below).
+
+Here are some more specific recommendations
+
+## Limit the number of concurrent streams per connection your protocol needs
```
BigLep, line 74: Can we give pointers on how to do this? 

For go-libp2p this means using resource manager right?

For rust, this isn't at the connection level, but somewhere I think we should be linking to https://docs.rs/libp2p/latest/libp2p/swarm/struct.SwarmBuilder.html#method.max_negotiating_inbound_streams .  Maybe we say, "rust-libp2p relies on each protocol to limit the number of streams per connection in XXX.  A global upperbound on negotiating/transient inbound streams can be set using https://docs.rs/libp2p/latest/libp2p/swarm/struct.SwarmBuilder.html#method.max_negotiating_inbound_streams."

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Fail2ban
+
+If you can log when a peer is misbehaving or is malicious, you can then hook up
+those logs to fail2ban and have fail2ban manage your firewall to automatically
+block misbehaving nodes. go-libp2p includes some builtin support for this
+usecase. More details below.
+
+
+## Leverage the resoure manager to limit resource (go-libp2p only)
```
BigLep, line 120:
```suggestion
## Leverage the resoure manager to limit resource usage (go-libp2p only)
```

Also update TOC

BigLep (commented):
Thanks @MarcoPolo for all the work here.  I'm game to take one last look once comments are incorporated.

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,311 @@
...
+accounting abstraction that you can make use of in your own application.
+
+## Rate limiting incoming connections (go-libp2p only)
+
+Depending on your use case, it can help to limit the number of inbound
+connections. You can use go-libp2p's
+[ConnectionGater](https://pkg.go.dev/github.com/libp2p/go-libp2p-core/connmgr#ConnectionGater)
+and `InterceptAccept` for this. For a concrete example, take a look at how Prysm
+implements their (Connection
+Gater)[https://github.com/prysmaticlabs/prysm/blob/63a8690140c00ba6e3e4054cac3f38a5107b7fb2/beacon-chain/p2p/connection_gater.go#L43].
```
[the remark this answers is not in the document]
MarcoPolo: whoops it should be [label](link)

content/reference/dos-mitigation.md
MarcoPolo: I should be more clear.

Here I'm talking about limiting the number of concurrent streams you need by design of the protocol, as opposed to using an existing protocol and trying to limit the streams at the end. For example imagine a RPC style protocol whose procedures are async and often take a long time to return (say > 1min). Here are two ways you could implement it:

1. Open a stream for each RPC call, and keep that stream open until the rpc call returns.
2. Open a stream for the start of the call then close it. The remote side will open a new stream with the answer.

Assume you make a lot of concurrent calls, method 1 would result in a large number of concurrent and mostly inactive streams. Method 2 would result in a fewer number of concurrent streams, and thus lower memory footprint.

If you add a limit here of say 10 streams, then method 1 will mean you can only have 10 concurrent RPC calls, while method 2 would let you have a much larger number of concurrent RPC calls.

Does that make sense? I should rephrase this to focus on the fact this is about protocol design (the inception stage of a p2p application) not about the deployed stage.

BigLep: Got it - makes sense.  

Side: lets find/create a place to point to https://docs.rs/libp2p/latest/libp2p/swarm/struct.SwarmBuilder.html#method.max_negotiating_inbound_streams .  Maybe there's a section about transient/negotiating connections and resources and that those should guarded against too.  Go and Rust both have some protections here.

MarcoPolo:
@BigLep Thanks for the review! I appreciate the help in getting us some good docs.

I believe I addressed all the comments. Another review would be very much appreciated.

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,381 @@
...
+# Incorporating DOS mitigation from the start
+
+The general strategy is to use the minimum amount of resources as possible and
+make sure that there's no untrusted amplification mechanism (e.g. an untrusted
+node can force you to do 10x the work it does). A protocol-level reputation
+system can help (take a look at [GossipSub](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub) for inspiration) as well as
+logging misbehaving nodes and actioning those logs separately (see fail2ban
+below).
+
+Here are some more specific recommendations
```
BigLep, line 75:
```suggestion
logging misbehaving nodes and actioning those logs separately 
(see [fail2ban](#how-to-automate-blocking-with-fail2ban) below).

Below are some more specific recommendations.
```

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,381 @@
...
+
+## Limit the number of connections your application needs
+
+Each connection has a resource cost associated with it. A connection will
+usually represent a peer and a set of protocols with each their own resource
+usage. So limiting connections can have a leveraged effect on your resource
+usage.
+
+In go-libp2p the number of active connections is managed by the
```
BigLep, line 84: A few nitpics here:
1. We use connmgr and ConnManager
2. When we're hyperlinking, I think it's good to remove the ticks so it's clear that it's a hyperlink.  (see screenshot of rendering)
3. We do "ConnManager" no space and "Resource Manager" with space.
4. Maybe the first time we talk about Resource Manager here we make it a hyperlink?

image

BigLep (approved):
HI @MarcoPolo - good stuff.  A few last comments but feel free to ship once incorporated.  Good times!

content/reference/dos-mitigation.md
```diff
@@ -0,0 +1,381 @@
...
+accounting abstraction that you can make use of in your own application.
+
+## Rate limiting incoming connections (go-libp2p only)
+
+Depending on your use case, it can help to limit the number of inbound
+connections. You can use go-libp2p's
+[ConnectionGater](https://pkg.go.dev/github.com/libp2p/go-libp2p-core/connmgr#ConnectionGater)
+and `InterceptAccept` for this. For a concrete example, take a look at how Prysm
+implements their (Connection
+Gater)[https://github.com/prysmaticlabs/prysm/blob/63a8690140c00ba6e3e4054cac3f38a5107b7fb2/beacon-chain/p2p/connection_gater.go#L43].
```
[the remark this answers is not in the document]
MarcoPolo: It's actually a misuse of `(foo)[bar]` vs `[foo](bar)`

content/reference/dos-mitigation.md
MarcoPolo: > We do "ConnManager" no space and "Resource Manager" with space.

In go-libp2p-core they're called `ConnManager` and `ResourceManager`. Using `Conn Manager` feels weird and so does `ResourceManager` although happy to make one change or the other if you feel strongly.

> When we're hyperlinking, I think it's good to remove the ticks so it's clear that it's a hyperlink. (see screenshot of rendering)

I think our template should support the fixed width + hyperlink. I'll see if I can fix it.

MarcoPolo: > I think our template should support the fixed width + hyperlink. I'll see if I can fix it.

Fixed this by adding a `text-decoration: underline` css property.

Screen Shot 2022-08-15 at 4 25 20 PM

BigLep (approved):
Looks great - lets ship!

PR #164, opened 2022-08-23
Add Plausible to libp2p docs

DannyS03:
Smart analytics for docs

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -22,7 +22,8 @@
     {{end}}
 
     <script src="{{"js/jquery-2.x.min.js"| relURL}}{{ if $assetBusting }}?{{ now.Unix }}{{ end }}"></script>
-
+    <script defer data-domain="docs.libp2p.io" src="https://plausible.io/js/plausible.js"></script>
+    
     <style type="text/css">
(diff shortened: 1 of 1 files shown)


johnnymatthews (approved):

mxinden:
Will the connected data be available somewhere public @DannyS03?

DannyS03:
@mxinden https://plausible.io/docs.libp2p.io :)

DannyS03:
as it is public, we can add an analytics section to the README (also going to update it in general)

mxinden:
Thank you @DannyS03!

> as it is public, we can add an analytics section to the README (also going to update it in general)

Good idea.

marten-seemann:
I don't think this dashboard should be public, and we also shouldn't link it from anywhere.

mxinden:
Can you expand on your reasoning @marten-seemann?

My mindset is: If we are already tracking folks, we should make it public, as in everyone can benefit from the data, not just us.

marten-seemann:
There's a privacy cost to tracking website visits, even is the tool used is less invasive than Google's tracking. This cost is increased by orders of magnitude by making that data publicly available.

PR #165, opened 2022-08-26
docs: update DOS Mitigation links to monorepo

p-shahi:
Complete "Update links in dos-mitigation.md" subtask of libp2p/go-libp2p#1556

Note:
Some of the previous links used the `master` branch when linking to code at specific line numbers. The contents of the file were modified and the link no longer pointed to the right line of code
- ex: setting limits on system scope pointed to [limit_defaults.go#L342](https://github.com/libp2p/go-libp2p-resource-manager/blob/master/limit_defaults.go#L342)

Or linked to code not on the master branch but on a specific commit id

- ex: [24b27cc71b7a62340f90d5f057e705cf10d5690f/p2p/protocol/identify/id.go#L150](https://github.com/libp2p/go-libp2p/blob/24b27cc71b7a62340f90d5f057e705cf10d5690f/p2p/protocol/identify/id.go#L150)

To address this, the links now point to the latest tagged version `v0.22.0`

### DIFF
--- a/content/reference/dos-mitigation.md
+++ b/content/reference/dos-mitigation.md
@@ -88,7 +88,7 @@ connections, and try to keep the number of connections above the low watermark.
 You can protect certain connections with the
 [`.Protect`](https://pkg.go.dev/github.com/libp2p/go-libp2p/p2p/net/connmgr#BasicConnMgr.Protect)
 method. The `ConnManager` is in charge of pruning connections to stay below the
-defined high watermark, in contrast, the [Resource Manager](github.com/libp2p/go-libp2p-resource-manager/) represents a hard
+defined high watermark, in contrast, the [Resource Manager](https://github.com/libp2p/go-libp2p/tree/master/p2p/host/resource-manager) represents a hard
 limit where connections will fail to be created in the first place once we've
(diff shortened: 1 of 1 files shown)


content/reference/dos-mitigation.md
```diff
@@ -3,6 +3,8 @@ title: "DOS Mitigation"
 weight: 3
 ---
 
+><em> This document is up to date as of go-libp2p [v0.22.0](https://github.com/libp2p/go-libp2p/releases/tag/v0.22.0)</em>
```
MarcoPolo, line 6: Is the expectation that this doc will be updated on every release? That seems untenable and not a great use of time.

MarcoPolo (approved):

content/reference/dos-mitigation.md
p-shahi: I don't think this doc (and website) should follow the same release process as the codebase and agree that it's unreasonable to update this every release. Since we make API references in places, this note was to let the reader know that it's using v0.22
I can remove it if you think it's confusing.

p-shahi: I removed it

PR #173, opened 2022-08-30
Make sure tutorial will work for user

danicuki:
Sometimes the user is using a different library version. I tested this tutorial many times before finding the right working version. 
It's good to have the version explicit in the docs.

### DIFF
--- a/content/tutorials/getting-started/javascript.md
+++ b/content/tutorials/getting-started/javascript.md
@@ -267,7 +267,9 @@ libp2p has stopped
 ```
 
 Success! Our two peers are now communicating over a multiplexed, secure channel.  Sure, they can only say "ping", but it's a start!
-## Working version
+
+## Working versions
+
 This tutorial uses the following js-libp2p package versions:
 ```
     "@chainsafe/libp2p-noise": "^8.0.1",
--- a/content/tutorials/getting-started/javascript.md
+++ b/content/tutorials/getting-started/javascript.md
@@ -277,7 +277,8 @@ This tutorial was created and tested using these packages versions:
     "libp2p": "^0.38.0",
     "multiaddr": "^10.0.1"
 ```
-If you are using any different version of those above, make sure to change your code and adapt it to work for you.  
+
+> Modify your js-libp2p packages if you start to experience issues to those mentioned above.
 
 ## What's next?
 
--- a/content/tutorials/getting-started/javascript.md
+++ b/content/tutorials/getting-started/javascript.md
@@ -267,6 +267,17 @@ libp2p has stopped
 ```
 
 Success! Our two peers are now communicating over a multiplexed, secure channel.  Sure, they can only say "ping", but it's a start!
+## Working version
+This tutorial was created and tested using these packages versions:
+```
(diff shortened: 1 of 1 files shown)


content/tutorials/getting-started/javascript.md
```diff
@@ -267,6 +267,17 @@ libp2p has stopped
 ```
 
 Success! Our two peers are now communicating over a multiplexed, secure channel.  Sure, they can only say "ping", but it's a start!
+## Working version
```
DannyS03, line 270:
```suggestion
## Working versions

```

content/tutorials/getting-started/javascript.md
```diff
@@ -267,6 +267,17 @@ libp2p has stopped
...
+This tutorial was created and tested using these packages versions:
```
DannyS03, line 271:
```suggestion
This tutorial uses the following js-libp2p package versions:
```

content/tutorials/getting-started/javascript.md
```diff
@@ -267,6 +267,17 @@ libp2p has stopped
...
+```
+    "@chainsafe/libp2p-noise": "^8.0.1",
+    "@libp2p/mplex": "^5.1.0",
+    "@libp2p/tcp": "^3.0.4",
+    "@libp2p/websockets": "^3.0.2",
+    "libp2p": "^0.38.0",
+    "multiaddr": "^10.0.1"
+```
+If you are using any different version of those above, make sure to change your code and adapt it to work for you.
```
DannyS03, line 280:
```suggestion

> Modify your js-libp2p packages if you start to experience issues to those mentioned above.
```

DannyS03 (changes requested):
thanks @danicuki, added some suggestions and would also suggest that this is at the beginning of the tutorial with the config instructions

DannyS03:
@danicuki checking here if you're okay with the suggestions

danicuki:
yes! thanks!

DannyS03 (approved):

marten-seemann (commented):
Putting version numbers here is a guarantee that we'll have a horribly outdated tutorial in a few months.

DannyS03:
> Putting version numbers here is a guarantee that we'll have a horribly outdated tutorial in a few months.

we did mention that we'd setup jobs in the CI/CD to run the published tutorials which would notify of version changes or breaks. Or, if they are already running against the examples on each repo, we can pull in those updates

marten-seemann:
> we did mention that we'd setup jobs in the CI/CD to run the published tutorials which would notify of version changes or breaks.

A set of versions that work once won't break. That doesn't prevent them from becoming outdated though.

danicuki:
> 

I've added the version with the purpose of helping the developer to make it work. Because, I was one that suffered trying to make it work and had to discover the right version. You are right: doc gets outdated. The point is, it will get outdated even if we don't put any version in the tutorial. So we must keep it updated anyway. The version here has the purpose of making at least work, even when it is outdated :-)

DannyS03:
with the stale nature of this change request and in correspondence with: https://github.com/libp2p/docs/issues/235, lets close this.

PR #174, opened 2022-08-31
Create Guides section

### LINKED ISSUE #29: Improve Examples section
The current examples section just links to external sources for go, js and rust.

I think there's a big opportunity to take those external examples and rework them a bit, pull them into the site, and use them as teaching tools to walk people through libp2p use cases.

I think that keeping the examples in their own repos will probably work best, since that's the easiest place to keep them up-to-date and integrated with CI.

My tentative plan is to clone the example repos as part of the build process for this site, run them through some formatting, and embed the result into the site.

For this to work well, some work needs to be done on the example repos so that they all have enough narrative & framing around them, and I need to develop the formatting / build process.

DannyS03:
## Context

- Creates dedicated section for technical how-tos, configuration, examples under a "Guides" category
- We will eventually pull in examples for each implementation, putting one or two 101-type of examples in the sub-section, then linking to examples source available for each implementation

Part of #29

## Latest preview

Please view the latest Fleek preview [here](https://bafybeidsr57n5bm2bptpjzx3cpuaffubmr354f724oncugv4qm4k7lklxi.on.fleek.co/).


marten-seemann (commented):
image
Not sure if it belongs to this PR, but should Examples and Tutorials be at the same level?

DannyS03:
> Not sure if it belongs to this PR, but should Examples and Tutorials be at the same level?

Do you mean that one should be a subsection of the other?

marten-seemann:
> > Not sure if it belongs to this PR, but should Examples and Tutorials be at the same level?
> 
> Do you mean that one should be a subsection of the other?

The other way around actually. I was suggesting that Example and Tutorials should both be nested directly under Guides.

DannyS03:
> > > Not sure if it belongs to this PR, but should Examples and Tutorials be at the same level?
> > 
> > 
> > Do you mean that one should be a subsection of the other?
> 
> The other way around actually. I was suggesting that Example and Tutorials should both be nested directly under Guides.

Oh they are @marten-seemann. "Tutorials" looks like a sub-section as the toggle is pushing the positioning of the section. They would be positioned equally once "Examples" has example pages. 

Added a placeholder file to fix: 
image

DannyS03:
Will add this to the overall update to include the Guides section and request a review there.

PR #180, opened 2022-09-05
Add crypto channels concept doc

DannyS03:
## Context

- Adds concept doc on crypto channels
- Focus is on Noise and TLS [1.3]


DannyS03:
Based on the enhanced structure, will follow with targeted PRs to add material for Noise and TLS 1.3.

Issue #178, opened 2022-09-05
Suggestion: i18n the docs. Such as translate it into Chinese.


ChengYueJia:
Suggestion like the title.

DannyS03:
Hi @ChengYueJia, thanks for the suggestion. This is something planned and we'll provide an update shortly on internationalization of the docs

ChengYueJia:
Glad to hear it. @DannyS03 

Two more questions:
Has it been scheduled? Has the Chinese version finished?

Actually, I wanna translate it into Chinese.

If u guys need, I can help to do it.

ChengYueJia:
@p-shahi Ok, I'll prepare for the Chinese Version PRs.

ChengYueJia:
Hi, @p-shahi  @DannyS03  I've met a problem.  

At first, I wanna organize the i18n files in different directories categories by lang.
just like this one.
```
├── content
|   ├── en
|   |   ├── post-01.md
|   |   ├── post-02.md
|   ├── cn
|   |   ├── post-01.md
|   |   ├── post-02.md
```

Later I wanna to set the i18n config.
And found docs of hugo-theme-learn, https://learn.netlify.app/en/cont/i18n/
But it shows the different organization as below.
This is less organized than the former one.

<img width="193" alt="image" src="https://user-images.githubusercontent.com/108982045/189101533-cf809c32-cb3a-4f9e-8fa3-ca4a31504871.png">

p-shahi:
Hey @ChengYueJia I was misinformed when asking you to contribute docs in other languages. After speaking with the team we decided that it will be difficult to maintain documentation in any language besides English as we do not have to resources necessary to write and proofread them. This is not libp2p specific as we don't maintain any localized version for IPFS or Filecoin. Our plan is to use a TMS which would automate translations in the future. For now I retract my ask for help and ask that we pause this effort. Thanks

ChengYueJia:
@p-shahi It's fine. Maybe TMS is a batter idea. 

I've see the [ethereum docs](https://github.com/ethereum/ethereum-org-website) is using the crowding as the TMS. 

Waiting for team's i18n scheduled time.

p-shahi:
I'll close this issue then

PR #184, opened 2022-09-07
i18n-zh: translate index and instruction into zh(Simple-Chinese).

ChengYueJia:
related with #178


ChengYueJia:
oops. seem like the Hugo is not ready to i18n.

PR #185, opened 2022-09-07
Remove match function references

DannyS03:
## Context

Seems like functions are no longer available.

## Latest preview

Please view the latest Fleek preview [here](https://bafybeif3wzq3qzdynmhuam7cw7ok5pjxe4hpcvpd4i7y6uusdniwlvuudy.on.fleek.co/concepts/protocols/)

### DIFF
--- a/content/concepts/protocols.md
+++ b/content/concepts/protocols.md
@@ -105,8 +105,6 @@ all of the registered match functions. If any returns `true`, the associated han
 This gives you a lot of flexibility to do your own "fuzzy matching" and define whatever rules for protocol matching
 make sense for your application.
 
-If you'd like to concurrently support a range of numbered versions, you may want to use semantic versioning (aka [semver](https://semver.org)).
-
 ### Dialing a specific protocol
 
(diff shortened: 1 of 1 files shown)


content/concepts/protocols.md
```diff
@@ -105,16 +105,8 @@ all of the registered match functions. If any returns `true`, the associated han
 This gives you a lot of flexibility to do your own "fuzzy matching" and define whatever rules for protocol matching
 make sense for your application.
 
-#### Match using semver
-
 If you'd like to concurrently support a range of numbered versions, you may want to use semantic versioning (aka [semver](https://semver.org)).
```
marten-seemann, line 108: Can we remove this sentence as well? We've moved away from sem-versioning protocols, since in practice it hasn't proved feasible to distinguish backwards-compatible and backwards-incompatible protocol changes.

DannyS03: Ah I see, updated.

marten-seemann (approved):

Issue #187, opened 2022-09-08
Add WebTransport to transport documentation 


jennijuju:
[WebTransport transport is an experimental, but full functioning transport](https://github.com/libp2p/go-libp2p/issues/1717) that's coming  in [go-libp2p v0.23.0 ](https://github.com/libp2p/go-libp2p/issues/1690)

We should add it to https://docs.libp2p.io/concepts/transport/ with an experimental note. 

reference: [spec](https://github.com/libp2p/specs/blob/d98386b5063735a74644d255275d100528535f9e/webtransport/README.md
)

DannyS03:
Cool, had also added a section for WebTransport in the proposed structure update: https://github.com/libp2p/docs/pull/186

DannyS03:
QUIC content is also missing and a priority so let's use this as an opportunity to add both. The overall transport content is tracked in: https://github.com/libp2p/docs/issues/17, but let's create a separate issue for QUIC, similar to WebTransport (this was also the intent after adding the proposed sub-chapters [here](https://github.com/libp2p/docs/pull/186), to create focussed issues)

PR #191, opened 2022-09-14
Add content for QUIC

### LINKED ISSUE #189: Add QUIC to transport documentation
Follows: https://github.com/libp2p/docs/issues/187#issuecomment-1241627105
Content is missing: https://github.com/libp2p/docs/issues/17#issuecomment-1228658994

> In our intent to add content for WebTransport, there is no existing content on QUIC to bounce off of

DannyS03:
## Context

- Adds QUIC content to transport documentation
- Content would be moved to dedicated page once (or if) new structure is in place.
- resolves #189, #187

## Latest preview

Please view the latest Fleek preview [here](https://bafybeiht3rrdpy7ixej72ogt6tjgrqjvi4vxgbwuqfmqkzvvnusikcnpdy.on.fleek.co/concepts/transport/).


mxinden:
//CC @elenaf9 given that you own https://github.com/libp2p/rust-libp2p/issues/2883.

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+processes running on the end systems). The purpose of newer transport protocols is not 
+only to improve current transport methods but also to allow for efficient connections 
+that distributed and peer-to-peer network stacks can utilize, like libp2p.
+
+{{% /notice %}}
+
+We need a transport protocol that:
+
+- Understands streams 
+- Is not byte-ordered
```
MarcoPolo, line 142: what is this?

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+cryptography methods). The streams in libp2p map cleanly to QUIC packets.
+
+When a connection starts, peers will take their host key and create a self-signed CA 
+certificate. They then sign an intermediate chain using their self-signed CA and put it 
+as a certificate chain in the TLS handshake.
+
+At the end of the handshake, each peer knows the certificate of the other. The peer can 
+verify if the connection was established with the correct peer by looking up the first 
+CA certificate on the certificate chain, retreive the public key, and using it to calculate 
+the opposing peer ID. QUIC acts like a record layer with TLS 1.3  as the backend as TLS is
```
MarcoPolo, line 209: I don't understand what a record vs backend layer is

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+At the end of the handshake, each peer knows the certificate of the other. The peer can 
+verify if the connection was established with the correct peer by looking up the first 
+CA certificate on the certificate chain, retreive the public key, and using it to calculate 
+the opposing peer ID. QUIC acts like a record layer with TLS 1.3  as the backend as TLS is 
+responsible for all the cryptography.
+
+{{% notice "info" %}}
+
+To be clear, there is no additional security handshake and stream muxer need as QUIC provides
```
MarcoPolo, line 214:
```suggestion
To be clear, there is no additional security handshake and stream muxer needed as QUIC provides 
```

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+<!-- to add link to secure comm guide, and later to the specific doc that covers TLS -->
+
+libp2p only supports bidirectional streams and uses TLS 1.3 by default (but can use other
+cryptography methods). The streams in libp2p map cleanly to QUIC packets.
+
+When a connection starts, peers will take their host key and create a self-signed CA 
+certificate. They then sign an intermediate chain using their self-signed CA and put it 
+as a certificate chain in the TLS handshake.
+
+At the end of the handshake, each peer knows the certificate of the other. The peer can
```
MarcoPolo, line 206: I'd link to the TLS spec in libp2p/specs

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+```
+/ip4/127.0.0.1/udp/65432/quic/
+```
+
+In this section, we offered an overview of QUIC and how QUIC works in libp2p.
+
+{{% notice "tip" %}}
+
+For more details on QUIC, including its limitations 
+check out the following resources:
```
MarcoPolo, line 231: ?

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+{{% /notice %}}
+
+## WebTransport
+
+Another transport protocol under development at the IETF is WebTransport.
+WebTransport is a new specification that uses QUIC to offer an alternative to
+WebSockets. Instead, it can be considered WebSockets over QUIC by allowing
```
MarcoPolo, line 239:
```suggestion
WebSockets. It can be considered WebSockets over QUIC by allowing 
```

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+## WebTransport
+
+Another transport protocol under development at the IETF is WebTransport.
+WebTransport is a new specification that uses QUIC to offer an alternative to
+WebSockets. Instead, it can be considered WebSockets over QUIC by allowing 
+browsers to establish a stream-multiplexed and bidirectional 
+connection to servers. 
+
+The specification can depend on and reuse the QUIC infrastructure in place 
+to offer WebSockets that take the benefits of UDP and offer sockets without head-of-line
```
MarcoPolo, line 244:
```suggestion
to offer WebSockets that take the benefits of UDP and offer streams without head-of-line 
```

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+WebSockets. Instead, it can be considered WebSockets over QUIC by allowing 
+browsers to establish a stream-multiplexed and bidirectional 
+connection to servers. 
+
+The specification can depend on and reuse the QUIC infrastructure in place 
+to offer WebSockets that take the benefits of UDP and offer sockets without head-of-line 
+blocking.
+
+Recall that WebSockets are bidirectional, full-duplex communication between two 
+points over a single-socket connection. WebTransports, as a result, can be used
```
MarcoPolo, line 248:
```suggestion
points over a single-socket connection. WebTransport can be used
```

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+like WebSockets, but with the support of multistreams.
```
MarcoPolo, line 249:
```suggestion
like WebSockets, but with the support of multiple streams.
```

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+WebTransport streams can be arbitrary in size and independent when possible. 
+They are reliable but can be canceled when possible. The datagrams in a 
+WebTransport connections are MTU-sized and can be unreliable when possible.
```
MarcoPolo, line 253: I don't follow what "when possible" means. 

We don't support unreliable datagrams with webtransport

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+{{% notice "caution" %}}
+
+There is a functioning WebTransport implementation in go-libp2p that is part
```
MarcoPolo, line 257:
```suggestion
There is an experimental WebTransport transport in go-libp2p that is part 
```

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+WebTransport streams can be arbitrary in size and independent when possible. 
+They are reliable but can be canceled when possible. The datagrams in a 
+WebTransport connections are MTU-sized and can be unreliable when possible.
+
+{{% notice "caution" %}}
+
+There is a functioning WebTransport implementation in go-libp2p that is part 
+of the v0.23 release.
+
+The implementation should be used experimentally and is not recommended for
```
MarcoPolo, line 260: Combine with the prev paragraph.

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+protocol that fits well with a modular network design.
+
+For a standard WebSocket connection:
+
+- 1 RTT for DNS resolution
+- 1 RTT for TCP handshake
+- 1 RTT for TLS handshake
+- 1 RTT for WebSocket handshake
+- 1 RTT for Multistream Security handshake
+- 1 RTT for libp2p handshake
```
MarcoPolo, line 275: What's this?

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+Plenty of handshakes and roundtrips: 6 RTTs: 5 handshakes + 1 DNS resolution
+
+WebTransport running over QUIC only requires 4 RTTs, as:
+
+- 1 RTT for QUIC handshake
+- 1 RTT for WebTransport handshake
+- 2 RTT for libp2p handshake; one for multistream and one for the secure
```
MarcoPolo, line 283: QUIC provides the security, we don't need to add TLS or noise here (we don't double encrypt).

We _do_ need to authenticate the connection and do that with a Noise handshake (see webtransport spec)

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+by `/webtransport` and a list of multihashes of the node certificates that the server uses.
+
+For instance, for multiaddress `/ip4/127.0.0.1/udp/123/quic/webtransport/certhash/<hash1>`, 
+a standard local QUIC connection is defined up until and including `/quic.` 
+Then, `/webtransport/` runs over QUIC and the self-signed certificate hash that the 
+server will use to verify the connection.
+
+WebTransport requires an HTTPS URL to establish a WebTransport session - 
+e.g., `https://docs.libp2p.com/webtransport` and the multiaddresses use an HTTP URL
+instead. Since multiaddrs don't allow the encoding of URLs, the HTTP endpoint of a libp2p
```
MarcoPolo, line 299: > Since multiaddrs don't allow the encoding of URLs,

This isn't exactly true. We just haven't defined what it would look like. Multiaddrs are able to encode urls, we just haven't come to consensus on what a canonical representation here looks like.

content/concepts/transport.md
```diff
@@ -115,3 +115,189 @@ and some places in the codebase still use the "swarm" terminology.
...
+Then, `/webtransport/` runs over QUIC and the self-signed certificate hash that the 
+server will use to verify the connection.
+
+WebTransport requires an HTTPS URL to establish a WebTransport session - 
+e.g., `https://docs.libp2p.com/webtransport` and the multiaddresses use an HTTP URL
+instead. Since multiaddrs don't allow the encoding of URLs, the HTTP endpoint of a libp2p 
+WebTransport servers must be located at `/.well-known/libp2p-webtransport`.
+
+For instance, the WebTransport URL of a WebTransport server advertising 
+`/ip4/1.2.3.4/udp/1234/quic/webtransport/` would be `https://1.2.3.4:1234/.well-known/libp2p-webtransport?type=tls`.
```
MarcoPolo, line 303: I think it's `?type=noise` because this refers to the authentication handshake.

MarcoPolo (changes requested):

content/concepts/transport.md
DannyS03: Addressed

content/concepts/transport.md
DannyS03: Addressed

content/concepts/transport.md
DannyS03: Updated

content/concepts/transport.md
DannyS03: Addressed

content/concepts/transport.md
DannyS03: thanks, forgot the links but I also think this isn't that necessary. I can re-add the note if otherwise

content/concepts/transport.md
DannyS03: Updated

content/concepts/transport.md
DannyS03: Addressed

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
 {{% /notice %}}
 
 [definition_switch]: /reference/glossary/#switch
+
+## QUIC
+
+Transport protocols continue to improve transport methods in an attempt to alleviate 
+the shortcomings of the transport layer.
```
marten-seemann, line 122: I don’t understand this sentence. How about a quick summary of what QUIC is instead, something like “QUIC is a new transport protocol that provides an always-encrypted, stream-multiplexed connection built on top of UDP. It started as an experiment by Google in xxxx and was later standardized by the IETF <link to RFC 9000>”?

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
 [definition_switch]: /reference/glossary/#switch
+
+## QUIC
+
+Transport protocols continue to improve transport methods in an attempt to alleviate 
+the shortcomings of the transport layer. 
+
+{{% notice "note" %}}
+
+Recalling the purpose of transports
```
marten-seemann, line 126: This paragraph is technically correct, but I don’t like the framing as transports being the thing that introduce reliability.

How about listing the problems that existed with TCP that led to the development of QUIC, namely:
1. it’s a single byte stream exposed by the kernel, so streams layered on top of TCP experience HoL blocking (this would be a good place to explain HoLB)
2. ossification: because TCP is unencrypted, middleboxes have been reading (and modifying!) TCP header fields and might break in unexpected ways when they encounter anything they don’t like
3. the TCP 3-way handshake is inefficient, as it spends 1 RTT on verifying the client’s address (that’s what the 3-way handshake is there for)

content/concepts/transport.md
DannyS03: Updated the roundtrip list

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+- Understands streams 
+- Is not byte-ordered
+- Overcomes HOL blocking (Head-of-line blocking)
+- Overcomes the latency of connection setup
+- Overcomes the ossification risks of TCP
+
+and, ideally, keeps the guarantees of TCP.
+
+In 2014, a new transport protocol 
+called QUIC (which, at the time, stood for Quick UDP Internet Connections, but now
```
marten-seemann, line 149: Please don’t use the abbreviation any more, it would be nice if people forgot about this ;)

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+In 2014, a new transport protocol 
+called QUIC (which, at the time, stood for Quick UDP Internet Connections, but now 
+is only referred to as QUIC and does not use the original acronym) was launched as an 
+experiment on Google Chrome. It has since been refined and maintained by an official 
+working group under the IETF (Internet Engineering Task Force). 
+
+QUIC is a UDP-based multiplexed and secure transport. 
+The official standard is defined in [RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000).
+
+Being UDP-based, QUIC optimizes for speedy transmission, as opposed to the latency
```
marten-seemann, line 157: That’s not correct, a big part of QUIC’s performance gains are due to latency reductions (fewer RTTs during the handshake!). After that, it all depends on the congestion controller you’re using.
But I don’t think we need to say anything here.

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+QUIC is a UDP-based multiplexed and secure transport. 
+The official standard is defined in [RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000).
+
+Being UDP-based, QUIC optimizes for speedy transmission, as opposed to the latency 
+that exists in HTTP leveraging TLS over TCP.
+
+A web browser connection typically entails the following (TCP+TLS+HTTP/2):
+
+1. IP layer: the connection will run on the IP layer, which is responsible for 
+   packet routing.
```
marten-seemann, line 163: I think we can remove this. No need to talk to much about the IP layer, as it’s the same for TCP and QUIC.

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+A web browser connection typically entails the following (TCP+TLS+HTTP/2):
+
+1. IP layer: the connection will run on the IP layer, which is responsible for 
+   packet routing. 
+2. Transport layer: TCP then runs on top of the IP layer to provide a reliable 
+   byte stream.
+3. Secure communication layer: Secure communication (i.e., TLS) runs on TCP to 
+   encrypt the bytes.
+   - Negotiation (SYN-ACK) of encryption parameters for TLS. Standard TLS over
```
marten-seemann, line 168: There’s no SYN-ACK in TLS (conceptually, yes, but it’s not called that way).

Maybe say something like this: A TLS handshake is run on top of the TCP connection, establishing an encrypted and authenticated connection between the two sides. A typical TLS 1.3 handshake takes 1 RTT.

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+experiment on Google Chrome. It has since been refined and maintained by an official 
+working group under the IETF (Internet Engineering Task Force). 
+
+QUIC is a UDP-based multiplexed and secure transport. 
+The official standard is defined in [RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000).
+
+Being UDP-based, QUIC optimizes for speedy transmission, as opposed to the latency 
+that exists in HTTP leveraging TLS over TCP.
+
+A web browser connection typically entails the following (TCP+TLS+HTTP/2):
```
marten-seemann, line 160: There’s a nice graphics floating around in various QUIC presentations showing QUIC’s position in the OSI stack. Maybe it would be worth including that one here?

content/concepts/transport.md
DannyS03: The idea is to show how QUIC takes the role of the record layer found in TLS

Updated

content/concepts/transport.md
DannyS03: Addressed

DannyS03:
Thanks @MarcoPolo! Suggestions and feedback incorporated

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+   encrypt the bytes.
+   - Negotiation (SYN-ACK) of encryption parameters for TLS. Standard TLS over 
+     TCP requires 3 RTT.
+4. Application layer: HTTP runs on a secure transport connection to transfer 
+   information.
+   - Data starts to flow.
+
+Secure TCP-based connections offer a multi-layered approach for secure communication 
+over IP, whereas QUIC, by design, is an optimized transport consolidation of layers. 
+QUIC has to deal with TCP-like congestion control, loss recovery, and encryption.
```
marten-seemann, line 176: Suggestion:
QUIC combines the functionality of these layers: it sends UDP packets. Therefore, it is responsible for loss detection and repair itself. The TLS 1.3 handshake is performed in the first flight, removing the 1 RTT cost of verifying the client’s address. And it exposes multiple streams (as opposed to just a single byte stream), so no stream multiplexer is needed at the application layer.

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+1. IP layer: the connection will run on the IP layer, which is responsible for 
+   packet routing. 
+2. Transport layer: TCP then runs on top of the IP layer to provide a reliable 
+   byte stream.
+3. Secure communication layer: Secure communication (i.e., TLS) runs on TCP to 
+   encrypt the bytes.
+   - Negotiation (SYN-ACK) of encryption parameters for TLS. Standard TLS over 
+     TCP requires 3 RTT.
+4. Application layer: HTTP runs on a secure transport connection to transfer
```
marten-seemann, line 170: Should we point out here that HTTP/2 applies a stream muxer?

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+Secure TCP-based connections offer a multi-layered approach for secure communication 
+over IP, whereas QUIC, by design, is an optimized transport consolidation of layers. 
+QUIC has to deal with TCP-like congestion control, loss recovery, and encryption. 
+Part of the application layer is also built directly into QUIC; when you
+run HTTP on top of QUIC; only a small shim layer exists that maps 
+[HTTP semantics](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html) 
+onto QUIC streams.
+
+To establish a connection, QUIC assumes that the node sends the right address over
```
marten-seemann, line 182: I think we can remove this paragraph. This is likely too specific.

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+the packet. QUIC saves one round-trip by doing this. If there is suspicion of an 
+attack, QUIC has a defense mechanism that can require a three-way handshake, but only 
+under particular conditions. A minimum packet size rule exists for the first packet to 
+ensure that small malicious packets like SYN packets cannot be sent and consume excessive 
+resources, as can be done with TCP.
+
+QUIC saves another round-trip in using TLS 1.3 by optimistically providing keyshares.
```
marten-seemann, line 189: That’s not done by providing keyshares. Suggestion:
QUIC allows the resumption of connections (0-RTT connections). In that case, the client send application right away, even before the QUIC handshake has finished.

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+If you have established a connection before, the host can send you a session ticket 
+that can be used to establish a new secure connection instantly, without any round-trips.
+
+> Over the last several years, the IETF has been working on a new version of TLS, TLS 1.3.
+> Learn more about TLS 1.3 and how it is used in libp2p on the secure communication concept guide.
+
+<!-- to add link to secure comm guide, and later to the specific doc that covers TLS -->
+
+libp2p only supports bidirectional streams and uses TLS 1.3 by default (but can use other
+cryptography methods). The streams in libp2p map cleanly to QUIC streams.
```
marten-seemann, line 199: There’s no way to use other handshakes.

content/concepts/transport.md
```diff
@@ -115,3 +115,190 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+When a connection starts, peers will take their host key and create a self-signed CA 
+certificate. They then sign an intermediate chain using their self-signed CA and put it 
+as a certificate chain in the TLS handshake. View the full TLS specification
+[here](https://github.com/libp2p/specs/blob/master/tls/tls.md).
+
+At the end of the handshake, each peer knows the certificate of the other. The peer can 
+verify if the connection was established with the correct peer by looking up the first 
+CA certificate on the certificate chain, retreive the public key, and using it to calculate 
+the opposing peer ID. QUIC acts like a record layer with TLS 1.3 as the backend as TLS is
```
marten-seemann: I think this paragraph can be removed, this is very specific to libp2p-tls and doesn’t matter unless you actually want to implement libp2p-tls (in which case you should use the spec).

marten-seemann (commented):
I did a first pass over the QUIC section. Haven’t touched WebTransport yet.

Maybe we can split this PR and first get QUIC merged, and then tackle WebTransport?

content/concepts/transport.md
DannyS03: Incorporated suggestions

content/concepts/transport.md
DannyS03: Incorporated suggestion

content/concepts/transport.md
DannyS03: That is true

content/concepts/transport.md
DannyS03: Addressed

content/concepts/transport.md
DannyS03: Gotcha:)

content/concepts/transport.md
DannyS03: nice suggestion, thought it would be good to also better define the flow and create subsection, with this being under its own

content/concepts/transport.md
DannyS03: Updated - combined the later paragraph with similar content which now introduces the section

content/concepts/transport.md
DannyS03: Updated

content/concepts/transport.md
DannyS03: Removed

content/concepts/transport.md
DannyS03: Removed

content/concepts/transport.md
DannyS03: Also removed

content/concepts/transport.md
DannyS03: Indeed, planning to add diagrams separately (across the concept docs being added) in separate PRs, but can add this one here as part of this (part of that is to also to avoid a bunch of migration with the assets with the new proposed structure)

DannyS03:
@marten-seemann thanks! Incorporated the suggestions and PRs now stacked

content/concepts/transport.md
```diff
@@ -115,3 +115,98 @@ and some places in the codebase still use the "swarm" terminology.
...
+[RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000).
+
+### Key challenges with TCP
+
+1. Head-of-line blocking (HoL blocking): TCP is a single byte stream exposed by the kernel, 
+   so streams layered on top of TCP experience HoL blocking.
+
+   {{% notice "info" %}}
+   HoL blocking occurs when a the head packet holds up the other packets in the transmission 
+   queue.
```
elenaf9, line 133:
```suggestion
   HoL blocking occurs when a lost, or out-of-order delivered head packet holds up the other packets in the transmission 
   queue. 
```

content/concepts/transport.md
```diff
@@ -115,3 +115,98 @@ and some places in the codebase still use the "swarm" terminology.
...
+   inspect, transform, and even manipulate network traffic. Examples of middleboxes are 
+   firewalls, NATs, proxies, and load balancers.
+   {{% /notice %}}
+
+3. Handshake inefficiency: the 3-way handshake is inefficient, as it spends 1-RTT on verifying 
+   the client’s address.
+
+   {{% notice "info" %}}
+   TCP requires a 3-way handshake to establish a reliable, bidirectional connection as it intends to.
+   {{% /notice %}}
```
elenaf9, line 150: (Personal opinion, others might disagree):
For both points (2. & 3.) I don't feel like the _notice_ adds much substance to what's already been said. In particular the notice on point 3 could be removed.

content/concepts/transport.md
```diff
@@ -115,3 +115,98 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+and, ideally, keeps the guarantees of TCP.
+
+A web browser connection typically entails the following (TCP+TLS+HTTP/2):
+
+1. Transport layer: TCP runs on top of the IP layer to provide a reliable 
+   byte stream.
+   - TCP provides a reliable, bidirectional connection between two end systems.
+2. Secure communication layer: A TLS handshake runs on top of TCP to,
+   establishing an encrypted and authenticated connection.
```
elenaf9, line 167:
```suggestion
2. Secure communication layer: A TLS handshake runs on top of TCP to
   establishing an encrypted and authenticated connection.
```

content/concepts/transport.md
```diff
@@ -115,3 +115,98 @@ and some places in the codebase still use the "swarm" terminology.
...
+   - Standard TLS over TCP requires 3-RTT. A typical TLS 1.3 handshake takes 1-RTT.
+3. Application layer: HTTP runs on a secure transport connection to transfer 
+   information and applies a stream muxer to serve multiple requests.
+   - Application data starts to flow.
+
+<!-- to add diagram -->
+
+### What is QUIC?
+
+QUIC combines the functionality of these layers: it sends UDP packets. Therefore,
```
elenaf9, line 177:
```suggestion
QUIC combines the functionality of these layers. It sends UDP packets and therefore
```
This phrasing is imo a bit misleading. I would either remove the colons or make the following points a bullet-point list.

content/concepts/transport.md
```diff
@@ -115,3 +115,98 @@ and some places in the codebase still use the "swarm" terminology.
...
+   information and applies a stream muxer to serve multiple requests.
+   - Application data starts to flow.
+
+<!-- to add diagram -->
+
+### What is QUIC?
+
+QUIC combines the functionality of these layers: it sends UDP packets. Therefore, 
+it is responsible for loss detection and repair itself. By using encryption, 
+QUIC avoid middleboxes. The TLS 1.3 handshake is performed in the first flight,
```
elenaf9, line 179:
```suggestion
QUIC avoids ossified middleboxes. The TLS 1.3 handshake is performed in the first flight, 
```

content/concepts/transport.md
```diff
@@ -115,3 +115,98 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+<!-- to add diagram -->
+
+### QUIC in libp2p
+
+libp2p only supports bidirectional streams and uses TLS 1.3 by default. 
+The streams in libp2p map cleanly to QUIC streams.
+
+When a connection starts, peers will take their host key and create a self-signed CA 
+certificate. They then sign an intermediate chain using their self-signed CA and put it
```
elenaf9, line 198: Unsure about this, but is it still a **CA** (= Certificate authority) certificate if it is self-signed? @marten-seemann?

content/concepts/transport.md
DannyS03: I agree here as well, for both notices you mention. Updated

content/concepts/transport.md
DannyS03: You can set up your own CA and use it to self-sign - my understanding is that as QUIC only works with secure connections, the certificate needs to be issued by a CA instead of it just being self-signed

marten-seemann: > the certificate needs to be issued by a CA instead of it just being self-signed

On the x509 level, that's the same thing. Any certificate that signs another certificates is a CA certificate. That said, browsers (and many other TLS implementaions) will (in their default setting) only accept CAs from a list of trusted CAs, either shipped with the OS or the browser.

QUIC doesn't introduce any special requirements here. The properties are exactly what they are for a normal TLS 1.3 handshake. QUIC just define a way to run a handshake that doesn't do a TLS 1.3 handshake.

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+QUIC is a new transport protocol that provides an always-encrypted, stream-multiplexed 
+connection built on top of UDP. It started as an experiment by Google on Google Chrome 
+in 2014, and was later standardized by the IETF in 
+[RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000).
+
+### Key challenges with TCP
+
+1. Head-of-line blocking (HoL blocking): TCP is a single byte stream exposed by the kernel, 
+   so streams layered on top of TCP experience HoL blocking.
```
marten-seemann, line 129:
```suggestion
   so streams layered on top of TCP experience head-of-line (HoL) blocking.
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+   {{% notice "info" %}}
+   HoL blocking occurs when a lost, or out-of-order delivered head packet holds up the other packets in the transmission 
+   queue.
```
marten-seemann, line 133:
```suggestion
   In TCP, head-of-line blocking occurs when a single packet is lost, and packets delivered after that need to wait in the kernel buffer until a retransmission for the lost packet is received.
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+1. Head-of-line blocking (HoL blocking): TCP is a single byte stream exposed by the kernel, 
+   so streams layered on top of TCP experience HoL blocking.
+
+   {{% notice "info" %}}
+   HoL blocking occurs when a lost, or out-of-order delivered head packet holds up the other packets in the transmission 
+   queue. 
+   {{% /notice %}}
+
+2. Ossification: Because TCP is unencrypted, middleboxes can inspect and modify
```
marten-seemann, line 136:
```suggestion
2. Ossification: Because the header of TCP packet is unencrypted, middleboxes can inspect and modify
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+   TCP header fields and may break unexpectedly when they encounter anything they don’t like.
```
marten-seemann, line 137:
```suggestion
   TCP header fields and may break unexpectedly when they encounter anything they don’t understand.
   This makes it practically impossible to deploy any changes to the TCP protocol that change the wire format.
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+3. Handshake inefficiency: the 3-way handshake is inefficient, as it spends 1-RTT on verifying 
+   the client’s address.
+
+We need a transport protocol that:
```
marten-seemann, line 142:
```suggestion
QUIC was designed with the following goals in mind:
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+- Understands streams
```
marten-seemann, line 144:
```suggestion
- Streams at the transport layer, thereby overcoming HoL blocking
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+2. Ossification: Because TCP is unencrypted, middleboxes can inspect and modify
+   TCP header fields and may break unexpectedly when they encounter anything they don’t like.
+
+3. Handshake inefficiency: the 3-way handshake is inefficient, as it spends 1-RTT on verifying 
+   the client’s address.
+
+We need a transport protocol that:
+
+- Understands streams 
+- Overcomes HOL blocking (Head-of-line blocking)
```
marten-seemann, line 145:
```suggestion
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+- Overcomes the latency of connection setup
```
marten-seemann, line 146:
```suggestion
- Reducing the latency of connection establishment to a single RTT for new connections, and to allow sending of 0-RTT application data for resumed connections
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+- Overcomes the ossification risks of TCP
```
marten-seemann, line 147:
```suggestion
- Encrypting as much as possible. This eliminates the ossification risk, as middleboxes aren't able to read any encrypted fields, and allows future evolution of the protocol
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+and, ideally, keeps the guarantees of TCP.
```
marten-seemann, line 149: Not sure which guarantees that would be.
```suggestion
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+   byte stream.
+   - TCP provides a reliable, bidirectional connection between two end systems.
+2. Secure communication layer: A TLS handshake runs on top of TCP to
+   establish an encrypted and authenticated connection.
+   - Standard TLS over TCP requires 3-RTT. A typical TLS 1.3 handshake takes 1-RTT.
+3. Application layer: HTTP runs on a secure transport connection to transfer 
+   information and applies a stream muxer to serve multiple requests.
+   - Application data starts to flow.
+
+<!-- to add diagram -->
```
marten-seemann, line 163: Can we do this in this PR?

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+We need a transport protocol that:
+
+- Understands streams 
+- Overcomes HOL blocking (Head-of-line blocking)
+- Overcomes the latency of connection setup
+- Overcomes the ossification risks of TCP
+
+and, ideally, keeps the guarantees of TCP.
+
```
marten-seemann, line 150: It would probably make sense to make this a section "Comparing HTTP/2 and HTTP/3"

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+QUIC supports the resumption of connections (0-RTT connections), allowing a 
+client to send application data right away, even before the QUIC handshake has finished.
+
+<!-- to add diagram -->
+
+### QUIC in libp2p
+
+libp2p only supports bidirectional streams and uses TLS 1.3 by default. 
+The streams in libp2p map cleanly to QUIC streams.
```
marten-seemann, line 185:
```suggestion
libp2p directly uses QUIC streams, without any additional framing.
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+When a connection starts, peers will take their host key and create a self-signed 
+CA certificate. They then sign an intermediate chain using their self-signed CA certificate 
+and put it as a certificate chain in the TLS handshake. View the full TLS specification
+[here](https://github.com/libp2p/specs/blob/master/tls/tls.md).
```
marten-seemann, line 190:
```suggestion
To authenticate each others' peer IDs, peers encode their peer ID into a self-signed certificate,
which they sign using their host's private key. This is the same way peer IDs are authenticated in
the [libp2p TLS handshake](https://github.com/libp2p/specs/blob/master/tls/tls.md).
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+{{% notice "note" %}}
+
+To be clear, there is no additional security handshake and stream muxer needed as QUIC 
+provides all of this by default.
```
marten-seemann, line 195:
```suggestion
To be clear, there is no additional security handshake and stream muxer needed as QUIC 
provides all of this by default. This also means that establishing a libp2p connection between
two nodes using QUIC only takes a single RTT.
```

content/concepts/transport.md
```diff
@@ -115,3 +115,88 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+{{% /notice %}}
+
+Following the multiaddress format described earlier, a standard QUIC connection will
+look like: `/ip4/127.0.0.1/udp/65432/quic/`.
+
+In this section, we offered an overview of QUIC and how QUIC works in libp2p.
```
marten-seemann, line 202:
```suggestion
```

marten-seemann (commented):
I think it would be valuable to add the OSI layer diagram in this PR. We'll probably have to word-smith the text accompanying that diagram a bit, but overall, I like where this PR is going :)

content/concepts/transport.md
```diff
@@ -115,3 +115,102 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+The IETF introduced a new hypertext transfer protocol standard in late 2018, 
+which turned into a proposed standard for HTTP/3 in 
+[RFC 9114](https://datatracker.ietf.org/doc/html/rfc9114). HTTP/3 combines the advantages 
+of the existing transfer protocols HTTP/2 and HTTP over QUIC in one standard for faster and 
+more stable data transmission.
+
+The following diagram illustrates the OSI model for HTTP/2 and HTTP/3:
+
+![HTTP/2 & HTTP/3 OSI model](https://cloudspoint.xyz/wp-content/uploads/2022/03/http3.png)
```
marten-seemann, line 163: This is the diagram I was looking for. Are we allowed to include it from a 3rd-party site without their permission?

content/concepts/transport.md
```diff
@@ -115,3 +115,102 @@ and some places in the codebase still use the "swarm" terminology.
...
+   information and applies a stream muxer to serve multiple requests.
+   - Application data starts to flow.
+
+In contrast, HTTP/3 runs over [QUIC](##what-is-quic), where QUIC is similar to 
+TCP+TLS+HTTP/2 and runs over UDP. Building on UDP allows HTTP/3 to bypass the challenges 
+found in TCP and use all the advantages of HTTP/2 and HTTP over QUIC.
+
+### What is QUIC?
+
+QUIC combines the functionality of these layers. It sends UDP packets and therefore
```
marten-seemann, line 183:
```suggestion
QUIC combines the functionality of these layers. Instead of TCP, it builds on UDP. When a UDP datagram is lost, it is not automatically retransmitted by the kernel. QUIC therefore takes responsibility...
```

content/concepts/transport.md
```diff
@@ -115,3 +115,102 @@ and some places in the codebase still use the "swarm" terminology.
...
+it is responsible for loss detection and repair itself. By using encryption, 
+QUIC avoids ossified middleboxes. The TLS 1.3 handshake is performed in the first flight, 
+removing the 1-RTT cost of verifying the client’s address. QUIC also exposes multiple
```
marten-seemann, line 186:
```suggestion
removing the cost of verifying the client’s address (saving an RTT). QUIC exposes streams (and not just a single byte stream)```

File: content/concepts/transport.md [hunk 55]
```diff
@@ -115,3 +115,102 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+QUIC combines the functionality of these layers. It sends UDP packets and therefore
+it is responsible for loss detection and repair itself. By using encryption, 
+QUIC avoids ossified middleboxes. The TLS 1.3 handshake is performed in the first flight, 
+removing the 1-RTT cost of verifying the client’s address. QUIC also exposes multiple 
+streams, so no stream multiplexer is needed at the application layer. Part of the application 
+layer is also built directly into QUIC; when you run HTTP on top of QUIC; only a small shim 
+layer exists that maps 
+[HTTP semantics](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html) 
+onto QUIC streams.
```
marten-seemann, line 191: I don't think that's accurate. We can probably just remove this last sentence.

content/concepts/transport.md
```diff
@@ -115,3 +115,102 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+QUIC supports the resumption of connections (0-RTT connections), allowing a
```
marten-seemann, line 193: Getting really pedantic, here, but there's a difference between resumption and 0-RTT (which we don't need to explain here) ;)
```suggestion
When a client has communicated with a certain server in the past, it can make use of QUIC's 0-RTT feature for subsequent connections. It can then send (encrypted) application data even before the QUIC handshake has finished.
```

content/concepts/transport.md
```diff
@@ -115,3 +115,102 @@ and some places in the codebase still use the "swarm" terminology.
...
+streams, so no stream multiplexer is needed at the application layer. Part of the application 
+layer is also built directly into QUIC; when you run HTTP on top of QUIC; only a small shim 
+layer exists that maps 
+[HTTP semantics](https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html) 
+onto QUIC streams.
+
+QUIC supports the resumption of connections (0-RTT connections), allowing a 
+client to send application data right away, even before the QUIC handshake has finished.
+
+<!-- to add diagram -->
```
marten-seemann, line 196: Which diagram do you want to add here?

content/concepts/transport.md
```diff
@@ -115,3 +115,102 @@ and some places in the codebase still use the "swarm" terminology.
...
+
+### QUIC in libp2p
+
+libp2p only supports bidirectional streams and uses TLS 1.3 by default. 
+libp2p directly uses QUIC streams, without any additional framing.
```
marten-seemann, line 201:
```suggestion
Since QUIC already provides an encrypted, stream-multiplexed connection, libp2p directly uses QUIC streams, without any additional framing.
```

marten-seemann (commented):
This is beginning to look really good. A few more comments.

marten-seemann (commented):
Looks like there's still 2 unaddressed comments. Did you miss those @DannyS03?

content/concepts/transport.md
DannyS03: Oh I originally had a reference added, and added it back.

DannyS03: Would eventually add this as an SVG

DannyS03:
@marten-seemann checking if there are any other parts to reconsider

MarcoPolo (approved):

marten-seemann (approved):

PR #223, opened 2022-11-17
fix/readme-logo-and-test-fleek

DannyS03:


DannyS03:
Previous build completely removed from the PL Fleek account, so "libp2p-docs" will no longer run.

PR #227, opened 2022-11-17
fix-and-update/fix-doks-navigation

DannyS03:
## Context

- Addresses the navigation and sub-heading bugs after shipping the Doks migration.
- This raises a further opportunity to consider the best way to organize the content, establishing a general template/archetype for the articles we write, and considering the best keywords and phrases for SEO purposes (e.g. "What is"). We will continue updates in follow-up PRs.

## Latest preview

Please view the latest Fleek preview [here](https://bafybeif2nskn76tvr7cw4klfnvnkzxf3p6wbs2wv2fmd7efktygdyyq4nq.on.fleek.co/).


p-shahi (approved):
lgtm

Issue #250, opened 2022-11-23
Create and populate discovery & routing section


DannyS03:
## Done Criteria 
- [x] Initiate overview doc on peer discovery
- [x] Initiate protocol docs for:
   
   - [x] Rendezvous
   - [x] mDNS
   - [ ] GossipSub
   - [ ] Kad-DHT

## Why Important
- Provides material on how peers discover and announce services to other peers.
- Documents protocols not yet documented.

## Notes
Populating the docs to be tracked in other issues.

Would follow a structure similar to the other sections, such as:
- What is Peer Discovery
- Rendezvous
- mDNS
- GossipSub
- Kad-DHT

PR #259, opened 2022-12-05
feat: rendezvous

DannyS03:
## Context

- Adds content for libp2p rendezvous
- Part of https://github.com/libp2p/docs/issues/250

### Notes

Section this document lives under will become populated with corresponding PRs.

## Latest preview

Please view the latest Fleek preview [here](https://bafybeihfppfn7gkue55rgw3eqphtd7w727pd2k4qn6pg6lzwsjtjiwkl6u.on.fleek.co/concepts/discovery/rendezvous/).


content/concepts/discovery/rendezvous.md
```diff
@@ -0,0 +1,66 @@
...
+responsibility of relaying packets between other nodes.
+
+## Rendezvous in libp2p
+
+{{< alert icon="💡" context="info" text="The current rendezvous implementation replaces the initial ws-star-rendezvous implementation with rendezvous daemons and a fleet of p2p-circuit relays." />}}
+
+The libp2p rendezvous protocol can be used for different use cases. It is used
+during bootstrap to discover circuit relays that provide connectivity for browser
+nodes. Generally, a peer can use known rendezvous points to find peers that provide
+critical network services. Rendezvous is also used throughout the lifetime of
```
mxinden, line 25:
```suggestion
network services. Rendezvous is also used throughout the lifetime of
```

Also non-critical services.

content/concepts/discovery/rendezvous.md
```diff
@@ -0,0 +1,66 @@
...
+an application for real-time peer discovery by registering and polling rendezvous points
```
mxinden, line 26: What is "realy-time" peer discovery?

content/concepts/discovery/rendezvous.md
```diff
@@ -0,0 +1,66 @@
...
+There is a default peer registration lifetime of 2 hours. Peers can optionally specify the
+lifetime using a TTL parameter in the `REGISTER` message, with an upper bound of 72 hours.
+
+The rendezvous protocol runs over libp2p streams using the protocol ID `/rendezvous/1.0.0`.
+
+<!-- TO ADD: Interaction diagrams and context -->
+
+### Rendezvous and publish-subscribe
+
+For effective real-time discovery, rendezvous can be combined with [libp2p publish/subscribe](../messaging/pubsub/overview). At a basic level, rendezvous can bootstrap pubsub by discovering peers
```
mxinden, line 56: To me "real-time" serves a different meaning and is used in e.g. video-conferencing or other use-cases where latency is directly noticeable by a user. I don't think that is the case for rendezvous.

content/concepts/discovery/rendezvous.md
```diff
@@ -0,0 +1,66 @@
...
+subscribed to a topic. The rendezvous would be responsible for publishing packets, subscribing,
+or unsubscribing from packet shapes.
+
+Pubsub can also be used as a mechanism for building rendezvous services, where a number
+of rendezvous points can federate using pubsub for internal real-time distribution while still
+providing a simple interface to clients.
```
mxinden, line 62: Is that implemented anywhere?

mxinden (commented):
Thanks for the write up @DannyS03. A couple of comments.

Maybe @thomaseizinger, as the author of the rust-libp2p implementation and specification, can you take a look?

thomaseizinger (commented):
Thanks for the ping @mxinden.

Before I comment on any details of the text, I'd like to point out that to my knowledge and compared to other discovery protocols like the DHT, rendezvous is not actually used by IPFS or other large deployment of libp2p.

For example, the implementation of the protocol in Go lives outside the main repository: https://github.com/berty/go-libp2p-rendezvous.
A search for "rendezvous" in the `go-libp2p` repository only yields (slightly misleading) examples: https://github.com/search?q=repo%3Alibp2p%2Fgo-libp2p+rendezvous+language%3AGo&type=code&l=Go

I am not sure what the scope of this recent docs update is but in my opinion, it would be better to:

- Update the rendezvous spec to remove the sections about "replacing ws-star-rendezvous" and other things that have never been implemented.
- Overall revise the spec to more clearly state what it is without any of the "we may do X, Y or Z" with this.
- Trim this document down to a high-level introduction
- Link to the spec for any details

Compared to the similar discussion we had in the [TLS PR](https://github.com/libp2p/docs/pull/261#pullrequestreview-1209224784), we actually _did_ invent this. Consequently, there isn't much we can say here in regards to "This is the general rendezvous protocol and libp2p uses only this bit". Instead, I think we should just give a brief introduction and then link to the spec for any details. If we give the spec a slight polish, it should be easier to approach which would be a net-win because it also benefits people that discover the spec from somewhere else other than this docs site :)

My 2c. Let me know what you guys think!

marten-seemann (changes requested):
I agree with the points made about the use of "real-time".

I also think we need to point out that rendezvouz is **not** decentralized. It's federated at best. That's not necessarily a bad thing, there are valid use cases for that, but it introduces one central point of failure into the network. This description can then be used to contrast rendezvouz with fully decentralized solutions like the DHT and Gossipsub.

It would also be valuable to point out that rendezvouz is _not_ used by IPFS and Filecoin (where is it used at all?).

thomaseizinger:
> It would also be valuable to point out that rendezvouz is _not_ used by IPFS and Filecoin (where is it used at all?).

We used it at my old job for a project which I think is still live (that is why I initially wrote the Rust implementation).

`nim-libp2p` has an [implementation](https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/protocols/rendezvous.nim) but I don't know about any deployments. (cc @Menduist)

The go implementation I linked above is from berty but I heard that is a forked version and thus may not be compatible? (cc @gfanton)

gfanton:
We use a fork of an implementation made by https://github.com/vyzo (https://github.com/libp2p/go-libp2p-rendezvous/tree/implement-spec) at Berty, and it has been working well for us. I'm not sure how compatible it is with the `nim-libp2p` implementation though. The main change we made was to improve the `Subscribe` method, which allowed us to add MQTT servers for real-time discovery notification.

As @Menduist mentioned, the Rendezvous service is especially useful on mobile devices where applications have short-lived sessions and cannot use the DHT efficiently due to factors such as bootstrap time and discovery latency. We created a multi-discovery system that advertises on both the DHT and the rendezvous to provide a fallback option, but I'm not completely satisfied with it because it relies too much on the rendezvous and doesn't act as a truly decentralized service. Additionally, the DHT is essentially meaningless in this setup because it is hidden by the rendezvous service and uses too much energy on mobile devices.

> Happy to be corrected here :) We'll try to do some simulations in the upcoming months to see if we can get to a fully decentralized rendez-vous giving better perfs than DHT for service discovery.

Please keep us in touch, I'm really looking forward to this 😄

content/concepts/discovery/rendezvous.md
DannyS03: Not aware of an implementation myself, but it is possible (as per the [spec](https://github.com/libp2p/specs/blob/master/rendezvous/README.md/#rendezvous-and-pubsub)).

mxinden (approved):
This is good to go from my end @p-shahi. What do you suggest as next steps? I don't understand the branching strategy here.

PR #260, opened 2022-12-05
feat: mdns

DannyS03:
## Context

- Adds content for libp2p mdns
- Part of https://github.com/libp2p/docs/issues/250

### Notes

- Section this document lives under will become populated with corresponding PRs
- Discovery section initiated in https://github.com/libp2p/docs/pull/259

## Latest preview

Please view the latest Fleek preview [here](https://bafybeihrghtydaifp3l5fn3mjkyduvyegvqzngq6gyj2g5662vzo3kiazq.on.fleek.co/). (preview outdated as not against master anymore)


p-shahi:
@jxs  can you help review this?

(Took liberty of unassigning Marten to help spread the load.)

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,51 @@
...
+
+MDNS, or multicast Domain Name System, uses a multicast system of DNS records over a
+local network. Nodes broadcast topics they're interested in into the network instead of
+querying a central name server. The discovery is limited to the peers in the local network.
+
+MDNS uses UDP to enable hostname resolution on local networks, allowing it to be used
+on networks that support multicast, such as an office. In turn, it can also use DNS to
+resolve hostnames to IP addresses on the Internet. MDNS does not require a central DNS
+server to operate, as a node can resolve hostnames to IP addresses in a decentralized
+manner.
```
marten-seemann, line 17: This is not really correct. We're not interested in hostname resolution, we're interested in service discovery. I don't think it has anything to do with DNS over the internet.

I'd go with something along these lines (taken from ChatGPT):
```suggestion
mDNS is short for multicast Domain Name System. It is a way for devices on a local network to discover and communicate with each other without the need for a central server. It is commonly used on home networks to allow devices such as computers, printers, and smart TVs to find each other and connect. It uses a protocol called multicast to broadcast messages on the network, allowing devices to discover each other and exchange information.
```

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,51 @@
...
+
+## MDNS in libp2p
```
marten-seemann, line 19: wrong capitalization

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,51 @@
...
+
+In libp2p, mDNS can be used as part of the [rendezvous protocol](rendezvous) to enable peers
+to discover and connect. When a peer wants to connect to another peer in a libp2p network,
+it can use mDNS to resolve the hostname of the other peer to its multiaddress.
```
marten-seemann, line 23: This is not correct. There's no rendezvouz server here. mDNS is used to find _all_ nodes on a local network, not to find a specific node.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,51 @@
...
+
+MDNS follows a request-response model, where a peer broadcasts a query request to *find all peers*
```
marten-seemann, line 25: I wouldn't call this request-response.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,51 @@
...
+of discovered peers. The response message is in the form of DNS record:
+
+`<service-name> PTR <peer-name>.<service-name>`,
+
+where `<service-name>` is the name of the service that is being advertised, and
+`<peer-name>` is the name of the peer [that is providing the service].
+The `<service-name>` part of the record indicates the domain in which the service
+and peer are located.
+
+As responses are received, the peer adds the other peers' information into its local database of peers.
```
marten-seemann, line 36: That's an implementation detail. We should only say that we encode the multiaddr of the peer into the DNS record.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,51 @@
...
+to discover and connect. When a peer wants to connect to another peer in a libp2p network,
+it can use mDNS to resolve the hostname of the other peer to its multiaddress.
+
+MDNS follows a request-response model, where a peer broadcasts a query request to *find all peers*
+on a local network to receive DNS response messages from peers which contain the peer information
+of discovered peers. The response message is in the form of DNS record:
+
+`<service-name> PTR <peer-name>.<service-name>`,
+
+where `<service-name>` is the name of the service that is being advertised, and
```
marten-seemann, line 31: It's `_p2p._udp.local`.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,51 @@
...
+`<peer-name>` is the name of the peer [that is providing the service].
```
marten-seemann, line 32: It's not. We don't have any use for that field, so we just fill it with a string with random characters.

marten-seemann (changes requested):
It would probably be useful if you could run a kubo node and inspect the mDNS query and response using wireshark, and add (the relevant parts) to this document.

marten-seemann:
> (Took liberty of unassigning Marten to help spread the load.)

@p-shahi Thank you, but I had already started my review, so I'm posting it anyway.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,47 @@
...
+of DNS records over a local network to discover and communicate with nodes. Nodes
+broadcast topics they're interested in instead of querying a central name server.
+The discovery, however, is limited to the peers in the local network. mDNS is commonly
+used on home networks to allow devices such as computers, printers, and smart TVs to
+find each other and connect. It uses a protocol called multicast to broadcast messages
+on the network, allowing devices to discover each other and exchange information.
+
+## mDNS in libp2p
+
+In libp2p, mDNS is used for service discovery.
```
marten-seemann, line 19: Nope! It's used for peer discovery.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,47 @@
...
+When a peer wants to connect to another peer in a libp2p network,
+it can use mDNS to resolve the hostname of the other peer to its multiaddr.
```
marten-seemann, line 21: No, we don't use it for address resolution. We use it to find _all_ other peers in the network.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,47 @@
...
+A peer can broadcast a query request to *find all peers* on a local network to
+receive DNS response messages from peers, which contain the peer information
+of discovered peers. The response message is in the form of a DNS record:
+
+`<service-name> PTR <peer-name>.<service-name>`,
+
+where `<service-name>` is `_p2p._udp.local`, the name of the service that is being
+advertised.
+> `peer-name` is not used for anything and can be filled with a string with random characters.
```
marten-seemann, line 30: This needs to be rephrased. Example:
In libp2p, we don't have any use for `peer-name`. Since the wire format requires us to send something, we send a string of random ASCII characters.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,47 @@
...
+
+A TXT record contains the multiaddresses that the peer is listening on. A peer
```
marten-seemann, line 32: This needs to point out that this belongs to the query that is being sent.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,47 @@
...
+encodes the multiaddr of the other peer into the DNS record. Each multiaddress is a TXT
```
marten-seemann, line 33: No, how would it know the multiaddr of other peers?

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,47 @@
...
+attribute with the form `dnsaddr=/.../p2p/QmId`.
+> `dnsaddr` is a protocol that instructs the resolver to look up multiaddr(s) in DNS TXT records for the
```
marten-seemann, line 35: It's not really a protocol, it's an encoding.

content/concepts/discovery/mDNS.md
```diff
@@ -0,0 +1,47 @@
...
+> domain name in its value section. To resolve a `dnsaddr` multiaddr, the domain name in the value section
+> must first be prefixed with `_dnsaddr.`. Then the peer must make a DNS query to look up TXT records for the domain. Multiple dnsaddr attributes and/or TXT records are allowed.
```
marten-seemann, line 37: That's not correct. There's really no reason to use `/dns` addresses here, which seems to be what you're describing. Most commonly, we just multiaddresses containing IP addresses here. It's a local network after all.

marten-seemann (commented):
There seems to be some misunderstanding of how mDNS works and what we're trying to achieve.

@DannyS03, have you tried Wiresharking the mDNS traffic originating from an IPFS node, and inspecting the packets? I think this could be really instructive. I also think (as I mentioned in my last review) that we should add some of that Wireshark output here as examples.

content/concepts/discovery-routing/mDNS.md
```diff
@@ -0,0 +1,70 @@
+---
+title: "mDNS"
+description: "mDNS uses a multicast system of DNS records over a local network to enable peer discovery."
+weight: 224
+---
+
+## What is mDNS?
```
marten-seemann, line 7: I always liked when we started our articles with a sentence or two explaining where things fit into the libp2p stack.

Example: libp2p uses mDNS to discover other libp2p nodes on the same local network, using the same protocol that a computer would to find printers.

content/concepts/discovery-routing/mDNS.md
```diff
@@ -0,0 +1,70 @@
...
+broadcast topics they're interested in instead of querying a central name server.
+The discovery, however, is limited to the peers in the local network. mDNS is commonly
+used on home networks to allow devices such as computers, printers, and smart TVs to
+find each other and connect. It uses a protocol called multicast to broadcast messages
+on the network, allowing devices to discover each other and exchange information.
+
+## mDNS in libp2p
+
+In libp2p, mDNS is used for peer discovery, allowing peers to find and
+communicate with each other on the same local network without any prior configuration.
```
marten-seemann, line 20: mDNS just does the finding part, transports do the communicating part.

content/concepts/discovery-routing/mDNS.md
```diff
@@ -0,0 +1,70 @@
...
+DNS message containing their discovery details. These details are stored in the additional
+records of the response and include the multiaddresses that the peer is listening on, as
+well as other information such as the peer's `<peer-name>` and `<host-name>`.
+
+The response message (that is, the answer to the DNS message) is in the form of a
+DNS record: `<service-name> PTR <peer-name>.<service-name>`. Additional record in the
+response message contains a peer's discovery details in the following form:
+`<peer-name>.<service-name> TXT "dnsaddr=..."`
+
+- `<peer-name>` is a unique case-insensitive identifier for the peer, although it is not
```
marten-seemann, line 35:
```suggestion
- `<peer-name>` is supposed to be a unique case-insensitive identifier for the peer, required by the mDNS protocol. However, it doesn't serve any meaningful..
```

content/concepts/discovery-routing/mDNS.md
```diff
@@ -0,0 +1,70 @@
...
+
+| No. | Time       | Source IP                | Destination IP         | Protocol | Length | Info                      |
+|-----|------------|--------------------------|------------------------|----------|--------|---------------------------|
+| 1   | 0.000000   | 10.0.0.25                | 224.0.0.251            | MDNS     | 75     | Standard query 0x7d14 PTR _p2p._udp.local, "QM" question |
+| 2   | 0.000130   | fe80::8d9:f02f:717f:2c03 | ff02::fb               | MDNS     | 95     | Standard query 0x7d14 PTR _p2p._udp.local, "QM" question |
+| 3   | 0.002497   | 10.0.0.25                | 224.0.0.251            | MDNS     | 918    | Standard query response 0x7d14 PTR vmvokx8c0b7rcjtx4e14iyaxw7otg05xyg954y5fnh1a9rderxtxycicgtj3nn._p2p._udp.local SRV 0 0 4001 vmvokx8c0b7rcjtx4e14iyaxw7otg05xyg954y5fnh1a9rderxtxycicgtj3nn.local TXT A 127.0.0.1 AAAA ::1 |
+| 4   | 0.002507   | fe80::8d9:f02f:717f:2c03 | ff02::fb               | MDNS     | 938    | Standard query response 0x7d14 PTR vmvokx8c0b7rcjtx4e14iyaxw7otg05xyg954y5fnh1a9rderxtxycicgtj3nn._p2p._udp.local SRV 0 0 4001 vmvokx8c0b7rcjtx4e14iyaxw7otg05xyg954y5fnh1a9rderxtxycicgtj3nn.local TXT A 127.0.0.1 AAAA ::1 |
```
marten-seemann, line 59: This doesn't show the multiaddr, does it?

content/concepts/discovery-routing/mDNS.md
```diff
@@ -0,0 +1,70 @@
...
+
+## mDNS in libp2p
+
+In libp2p, mDNS is used for peer discovery, allowing peers to find and
+communicate with each other on the same local network without any prior configuration.
+This is achieved through multicast DNS (mDNS) records, which are sent to all nodes on the
+local network.
+
+To initiate peer discovery, a peer sends a query to all other peers on the network using
+a DNS message with the question `_p2p._udp.local PTR`. In response, each peer will send a
```
marten-seemann, line 25: The question is for `_p2p` (or is it `p2p`?) The rest is just there to make things fit into mDNS.

content/concepts/discovery-routing/mDNS.md
```diff
@@ -0,0 +1,70 @@
...
+on the network, allowing devices to discover each other and exchange information.
+
+## mDNS in libp2p
+
+In libp2p, mDNS is used for peer discovery, allowing peers to find and
+communicate with each other on the same local network without any prior configuration.
+This is achieved through multicast DNS (mDNS) records, which are sent to all nodes on the
+local network.
+
+To initiate peer discovery, a peer sends a query to all other peers on the network using
```
marten-seemann, line 24: We should say that this is using multicast, and therefore:
1. only works on local networks
2. doesn't require the peer to know any IP addresses of other nodes

content/concepts/discovery-routing/mDNS.md
marten-seemann: Thinking about it, I'm not sure if we need to mention this at all. This seems to be a bit too low-level for documentation. wdyt?

content/concepts/discovery-routing/mDNS.md
MarcoPolo: it's for _p2p._udp.local.
Screenshot 2023-01-26 at 4 37 04 PM

I think this is to low level. If I need to use wireshark to verify this, I don't think users will care about the specifics here.

content/concepts/discovery-routing/mDNS.md
MarcoPolo: If we're going to show this, we may as well show the multiaddr result of the query.

content/concepts/discovery-routing/mDNS.md
MarcoPolo, line 24: I would suggest getting rid of everything below here until the `Example` section. That one is kind of cool to see. Maybe add a bit more context for the example since we've removed this.

MarcoPolo (changes requested):
Tapping marten out, and I'll review from here on out.

content/concepts/discovery-routing/mDNS.md
marten-seemann: Yes, that was my motivation when I asked for including a Wireshark capture. 

Maybe this could even be a screenshot of Wireshark? No strong opinion in either direction though.

content/concepts/discovery-routing/mDNS.md
```diff
@@ -0,0 +1,47 @@
...
+used on home networks to allow devices such as computers, printers, and smart TVs to
+find each other and connect. It uses a protocol called multicast to broadcast messages
+on the network, allowing devices to discover each other and exchange information.
+
+## mDNS in libp2p
+
+In libp2p, mDNS is used for peer discovery, allowing peers to find and
+communicate with each other on the same local network without any prior configuration.
+This is achieved through multicast DNS (mDNS) records, which are sent to all nodes on the
+local network.
```
p-shahi, line 22:
```suggestion
```
Unnecessary sentence

content/concepts/discovery-routing/mDNS.md
p-shahi: Let's not include this? Why have wireshark output for mDNS and not for other protocols?

I'm in favor of removing this and the following paragraphs.

Instead we can describe the peer discover section https://github.com/libp2p/specs/blob/master/discovery/mdns.md#peer-discovery in two sentences and point to the spec

p-shahi (approved):

content/concepts/discovery-routing/mDNS.md
p-shahi: @MarcoPolo @marten-seemann please re-review. I removed the wireshark section and instead linked to the specs

MarcoPolo (approved):
@p-shahi could you please give this a once over before merging. ty!

PR #280, opened 2023-01-05
feat: noise

### LINKED ISSUE #19: Populate secure communication section: Noise & TLS
We should cover how libp2p's support for encrypted communication works, from the negotiation process (link to multistream concept article) to the specific implementations (~secio~, tls, etc).

## Done Criteria
- [x] Enhance secure comm overview
- [x] Populate Noise doc
- [x] Populate TLS doc

## Why Important
No explicit material on the supported secure channels in libp2p.

DannyS03:
## Context

- populates the noise document, similar to [TLS](https://docs.libp2p.io/concepts/secure-comm/tls/)
- closes #19

## Latest preview

Please view the latest Fleek preview [here](https://bafybeieqnt2njw3m73ttumxuchoo53pwgowf2h5xxghehr3fyqyo2lbruy.on.fleek.co/concepts/secure-comm/noise/).

### DIFF
--- a/content/concepts/secure-comm/noise.md
+++ b/content/concepts/secure-comm/noise.md
@@ -19,19 +19,20 @@ to create a secure protocol.
 
 noise-libp2p is an implementation of the Noise Protocol Framework used to establish a
 secure channel between two peers in the libp2p network. The protocol ID is
-`/noise`, and future versions of the specification may define new protocol IDs using the "/noise"
-prefix (e.g., `/noise/2`).
+`/noise`, and future versions of the specification may define new protocol IDs using the
+"/noise" prefix (e.g., `/noise/2`).
 
 When two peers connect, the transport upgrader negotiates
-which security and multiplexing protocols to use using
+which security and [multiplexing](../multiplex/overview.md) protocols to use using
 [multistream-select](https://github.com/multiformats/multistream-select).
 > A successor to multistream-select, called multiselect 2, is in development, but noise-libp2p is
 > compatible with the current upgrade process and multiselect 2.
 
 ### Handshake
 
-Peers exchange public keys and perform a Diffie-Hellman exchange to generate a pair of symmetric
-keys that can be used to encrypt traffic during the handshake.
+Peers exchange public keys and perform a
+[Diffie-Hellman exchange](https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange)
+to generate a pair of symmetric keys that can be used to encrypt traffic during the handshake.
 The static DH key used in the Noise protocol is authenticated using the libp2p identity keypair.
 
 The Noise Protocol Framework provides several different handshake patterns to choose from, each
@@ -46,8 +47,8 @@ additional handshake patterns may be added to noise-libp2p to support different 
 
 In addition to the standard Noise handshake, noise-libp2p also includes an extension registry, which
 allows for the exchange of additional data during the handshake process. This extension is used to
-exchange libp2p-specific data such as peer IDs and supported protocol versions that enable early
-stream muxer negotiation.
+exchange libp2p-specific data such as [peer IDs](../fundamentals/peers.md##peer-id) and supported
+protocol versions that enable [early stream muxer negotiation](../multiplex/early-negotiation).
 
 <!-- ADD DIAGRAM -->
 
--- a/content/concepts/secure-comm/noise.md
+++ b/content/concepts/secure-comm/noise.md
@@ -6,5 +6,58 @@ aliases:
     - "/concepts/secure-comm/noise"
 ---
 
-<!-- ADD NOTICE -->
-Coming soon!
+## What is Noise?
+
+The [Noise Protocol Framework](https://noiseprotocol.org/) is a set of tools for creating
+secure communication protocols by combining cryptographic primitives into patterns with
+verifiable security properties. It provides a high level of flexibility, allowing users to
+choose from a range of cryptographic primitives and design patterns to suit their needs.
+However, it is up to the user to carefully consider the tradeoffs and make informed decisions
+to create a secure protocol.
+
+## Noise in libp2p
+
+noise-libp2p is an implementation of the Noise Protocol Framework used to establish a
+secure channel between two peers in the libp2p network. The protocol ID is
+`/noise`, and future versions of the specification may define new protocol IDs using the "/noise"
+prefix (e.g., `/noise/2`).
+
+When two peers connect, the transport upgrader negotiates
+which security and multiplexing protocols to use using
+[multistream-select](https://github.com/multiformats/multistream-select).
+> A successor to multistream-select, called multiselect 2, is in development, but noise-libp2p is
+> compatible with the current upgrade process and multiselect 2.
+
+### Handshake
+
+Peers exchange public keys and perform a Diffie-Hellman exchange to generate a pair of symmetric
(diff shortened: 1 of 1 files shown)


content/concepts/secure-comm/noise.md
```diff
@@ -6,5 +6,59 @@ aliases:
...
-<!-- ADD NOTICE -->
-Coming soon!
+## What is Noise?
+
+The [Noise Protocol Framework](https://noiseprotocol.org/) is a set of tools for creating
+secure communication protocols by combining cryptographic primitives into patterns with
+verifiable security properties. It provides a high level of flexibility, allowing users to
+choose from a range of cryptographic primitives and design patterns to suit their needs.
+However, it is up to the user to carefully consider the tradeoffs and make informed decisions
+to create a secure protocol.
```
MarcoPolo, line 16:
```suggestion
verifiable security properties.
```

libp2p already has made these choices. Users here don't have to think about this.

content/concepts/secure-comm/noise.md
```diff
@@ -6,5 +6,59 @@ aliases:
...
+[Diffie-Hellman exchange](https://en.wikipedia.org/wiki/Diffie%E2%80%93Hellman_key_exchange)
+to generate a pair of symmetric keys that can be used to encrypt traffic during the handshake.
+The static DH key used in the Noise protocol is authenticated using the libp2p identity keypair.
+
+The Noise Protocol Framework provides several different handshake patterns to choose from, each
+with its tradeoffs in terms of security and performance. noise-libp2p currently supports the
+[`XX` pattern](https://noiseprotocol.org/noise.html#interactive-handshake-patterns-fundamental),
+which provides strong security guarantees but is slower than other options. In the future,
+additional handshake patterns may be added to noise-libp2p to support different use cases.
+
```
MarcoPolo, line 43: I'd cut this whole section.

```suggestion
```

content/concepts/secure-comm/noise.md
```diff
@@ -6,5 +6,59 @@ aliases:
...
+noise-libp2p is an implementation of the Noise Protocol Framework used to establish a
+secure channel between two peers in the libp2p network. The protocol ID is
+`/noise`, and future versions of the specification may define new protocol IDs using the
+"/noise" prefix (e.g., `/noise/2`).
+
+When two peers connect, the transport upgrader negotiates
+which security and [multiplexing](../multiplex/overview.md) protocols to use using
+[multistream-select](https://github.com/multiformats/multistream-select).
+> A successor to multistream-select, called multiselect 2, is in development, but noise-libp2p is
+> compatible with the current upgrade process and multiselect 2.
```
MarcoPolo, line 29: What is this?

Regardless, we don't need to mention multistream select here because the noise module takes an unencrypted data channel and returns an encrypted data channel. It doesn't know anything about multistream.

content/concepts/secure-comm/noise.md
```diff
@@ -6,5 +6,59 @@ aliases:
...
+additional handshake patterns may be added to noise-libp2p to support different use cases.
+
+<!-- ADD DIAGRAM -->
+
+#### The Noise Extension
+
+In addition to the standard Noise handshake, noise-libp2p also includes an extension registry, which
+allows for the exchange of additional data during the handshake process. This extension is used to
+exchange libp2p-specific data such as [peer IDs](../fundamentals/peers.md##peer-id) and supported
+protocol versions that enable [early stream muxer negotiation](../multiplex/early-negotiation).
```
MarcoPolo, line 51: Not relevant for the target audience (users of libp2p). We should cut.

content/concepts/secure-comm/noise.md
```diff
@@ -6,5 +6,59 @@ aliases:
...
+<!-- ADD DIAGRAM -->
+
+### Wire Format and Encryption
+
+After the Noise handshake is completed, the resulting encryption keys send ciphertext messages
+back and forth over the secure channel. The wire format for these messages and the cryptographic primitives
+used for encryption is specified in the Noise specification. noise-libp2p currently uses the
+[ChaCha20Poly1305](https://en.wikipedia.org/wiki/ChaCha20-Poly1305) AEAD cipher for encryption, which provides
+both confidentiality and integrity protection. The wire format consists of a header and a payload, with the
+payload being encrypted using the encryption keys derived from the Noise handshake.
```
MarcoPolo, line 62:
```suggestion
used for encryption is specified in the Noise specification.
```

MarcoPolo (changes requested):

DannyS03:
thanks for the review, @MarcoPolo!

I don't see the value in the docs if we abstract the material to the point where we are generally only mentioning something along the lines of "this is a protocol, this is the protocol in libp2p, and this is the protocol ID". However, I also agree that we should continue to abstract material that is primarily relevant to libp2p users.

What about an approach where we present more context on a subject matter in an *optional* dropdown menu. For example:

image

content/concepts/secure-comm/noise.md
DannyS03: This is mentioned because it is libp2p specific. We also describe it [here](https://docs.libp2p.io/concepts/multiplex/early-negotiation/#extension-registry-in-noise).

MarcoPolo:
> I don't see the value in the docs if we abstract the material to the point where we are generally only mentioning something along the lines of "this is a protocol, this is the protocol in libp2p, and this is the protocol ID". However, I also agree that we should continue to abstract material that is primarily relevant to libp2p users.

The most important things are:
1. What is this thing at a high level, and where can I learn more about this thing?
2. How does libp2p use this at a high level, and where can I learn more?
3. How does this help me.

For Noise this would be:
1. An encryption scheme that has been heavily reviewed and is used by Signal. learn more at https://noiseprotocol.org.
2. libp2p uses this as a way to encrypt data between nodes and provide forward secrecy. See more at the <link to spec>
3. This helps you send data end to end encrypted by default.

A user reading this doc should be able to understand how they would make use of this protocol. When I buy a hammer I only need to know which part I whack, not the process of making the hammer.

---

The problem with putting too much stuff in here is that then we need to review everything much more carefully. This turns the documentation into a spec rewrite. And makes it really brittle to any future changes. The higher level stuff targetting users is more likely to be true no matter what small changes we make. e.g. Noise will never _not_ be used to encrypt your data end to end.

content/concepts/secure-comm/noise.md
DannyS03: Incorporated

content/concepts/secure-comm/noise.md
DannyS03: Updated

DannyS03:
I hear you @MarcoPolo, thanks. I incorporated the suggestions.

content/concepts/secure-comm/noise.md
```diff
@@ -6,5 +6,27 @@ aliases:
...
 
-<!-- ADD NOTICE -->
-Coming soon!
+## What is Noise?
+
+The [Noise Protocol Framework](https://noiseprotocol.org/) is a widely-used encryption
+scheme that allows for secure communication by combining cryptographic primitives into
+patterns with verifiable security properties.
+
+Learn more at https://noiseprotocol.org.
```
DannyS03, line 15:
```suggestion
Learn more at [https://noiseprotocol.org](https://noiseprotocol.org).
```

MarcoPolo (approved):
Thank you :)

PR #285, opened 2023-01-09
feat: fetch

DannyS03:
## Context

Adds fetch as part of protocol basics.

## Latest preview

Fleek preview unavailable as PR not against master.

### DIFF
--- a/content/concepts/introduction/protocols/fetch.md
+++ b/content/concepts/introduction/protocols/fetch.md
@@ -4,4 +4,23 @@ description: ""
 weight: 24
 ---
 
-Coming soon!
+## Fetch in libp2p
+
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
I'm not familiar with this protocol. Looks like it's only used by PubSub? Should this be a section of the PubSub document then?

DannyS03:
> I'm not familiar with this protocol. Looks like it's only used by PubSub? Should this be a section of the PubSub document then?

Given the [potential effects](https://github.com/libp2p/specs/tree/master/fetch#potential-effects), seems like it also suits as a "general protocol".

mxinden:
In my eyes the fetch protocol is a legacy, not used anywhere. Thus I am closing here. Anyone, please reopen in case you think otherwise.

PR #192, opened 2022-09-16
feat: kaddht content routing doc

DannyS03:
## Context

- Adds initial Kademlia content routing piece
  > Some of this content may be a bit IPFS-specific, feedback would be great on this too
- resolves #23

## Latest preview

Please view the latest Fleek preview [here](https://bafybeifkqig6oey4ooqtkxbo2wh6o6l3ikwegufnvqut3hxqcnal7y7j3y.on.fleek.co/concepts/content-routing/).

PR #293, opened 2023-01-15
feat: kad-dht content routing

DannyS03:
## Context

- replaces and simplifies https://github.com/libp2p/docs/pull/192 
- aims to be **agnostic** from use cases (e.g. IPFS).

## Latest preview


p-shahi:
@Jorropo would you be able to help review this doc?

p-shahi:
That sounds good to me @mxinden I think we can close this, add to the IPFS docs if appropriate and then link to that document instead

Issue #236, opened 2022-11-20
Modularize and enhance Publish/Subscribe content


DannyS03:
## Done Criteria
- [ ] Create an introductory doc for publish/subscribe, similar to the other core component sections (e.g. [Transports](https://docs.libp2p.io/concepts/transports/overview/), [Steam Multiplexing](https://docs.libp2p.io/concepts/multiplex/overview/)). The title can be "What is Publish/Subscribe"
- [ ] Create an interface overview doc, similar to [Listening and Dialling](https://docs.libp2p.io/concepts/transports/listen-and-dial/) under Transports. The title can be "PubSub Wire Format"
- [ ] Create a floodsub doc.
  - Specify the purpose of floodsub and direct reader to gossipsub for production use cases.
- [ ] Create a gossipsub doc. Content updates are captured in https://github.com/libp2p/docs/issues/169.
- [ ] Add more examples as part of "example uses of pub/sub:"
- [ ] General writing style updates.
   - [ ] Gramatical.
   - [ ] There are descriptions that are currently commented out that were used as the image descriptors - these are useful descriptions that should be incorporated into the actual text on the page.
- [ ] Remove diagram backgrounds so that they are transparent.

## Why Important
While a solid explainer, there is opportunity to improve the current material and modularize the doc itself.

## Notes
-  The section would resemble the following structure:
    - What is Publish/Subscribe
    - PubSub Wire Format
    - FloodSub
    - GossipSub
- For now, we can update the images to have a transparent background so they fill the pages better (i.e. when in dark mode). However, we should eventually update all the pub/sub diagrams as part of the unified libp2p diagrams.

PR #299, opened 2023-01-17
feat: pubsub overview

DannyS03:
## Context

- Creates a common overview for Pub-Sub similar following a similar approach to the other sections
- Incorporates concepts from the original Pub-Sub doc
- part of #236

## Latest preview

Not against master (being worked on)


mxinden (approved):
Ready to merge from my end. @p-shahi can you merge once ready?

PR #334, opened 2023-07-23
fix(tutorial): correct syntax error and refine output in libp2p example

viniciusgomes:
- Added a missing comma after the `streamMuxers` property to keep proper object syntax.
- Removed the "libp2p has stopped" output for clearer console feedback during remote peer ping operations.

These changes aim to provide a smoother user experience and enhanced clarity when following the tutorial.


p-shahi (approved):
Ty!

PR #335, opened 2023-07-23
fix(docs): correct broken link for "peers" in the introduction

viniciusgomes:
Changed the link for "peers" from `/concepts/introduction/fundamentals/peers` to `/concepts/fundamentals/peers` to fix the redirection error.


p-shahi (approved):
Thank you!

PR #341, opened 2023-08-04
Enable link check CI

dchiquito:
* Exclude twitter links from CI check
* Enable link check worklow on any push/merge to master
* Fix typo in filecoin slack link

~~I don't have permission to dispatch the worklow so I guess this will be tested when it's merged to master :crossed_fingers:~~
Never mind, looks like it's run on attempted merges to master :+1:


dchiquito:
After some testing, it looks like this workflow only checks the source content directory, which contains unprocessed relative links, which are uncheckable by lychee. This means only links to outside resources are checked, but not any internal links.

I think the solution is to move link checking to the build job so it can run against the final build artifact.

PR #347, opened 2023-09-17
Update overview.md

postmeback:
Removed unnecessary back-slash


Stebalien:
Fixed in https://github.com/libp2p/docs/pull/348

PR #351, opened 2023-09-25
Update go.md

PanK0:
Added a specification to create and compile the second node with a different address.


PanK0:
Added a specification to create and compile the second node with a different address.

PR #352, opened 2023-09-25
Update go.md

PanK0:
Added the indication to write and compile the second node with another address


marten-seemann:
Listening on :0 picks a random different port. This is standard UNIX behavior.

PanK0:
I will check it.
Thanks.
I am sorry for this, I am just trying to learn something new.

Il Lun 25 Set 2023, 18:30 Marten Seemann ***@***.***> ha
scritto:

PR #358, opened 2023-12-14
Update javascript.md

elecbug:
Hello, I'm libp2p beginner used by js

but this mark down is have some invalid information as @libp2p/ping part...

maybe you are renamed package, I think.

thank you reading


Stebalien:
Fixed in https://github.com/libp2p/docs/pull/362 (I merged that one as it had a smaller diff)

PR #366, opened 2024-02-13
chore: update npm

Stebalien:
Get dependeabot to go away for a while.


Stebalien:
You know what? Now I remember why I hate JavaScript. Keeping up with the breaking changes just isn't worth it.

PR #367, opened 2024-02-13
fix: better clarity

junekhan:
I interpreted the story differently, as in, I thought there were four participants here: me, QmAlice, QmBob, and QmRelay. `I have a peer with the Peer ID QmAlice` meant `I have a friend with Peer ID QmAlice` to me in the first place.

Or the other way around, `I want to give out my address` can be changed to `I want to give out my address of QmAlice`, which may seem verbose.


Stebalien:
How about: "I have a libp2p node with a Peer ID...".

junekhan:
> How about: "I have a libp2p node with a Peer ID...".

@Stebalien Sounds good, that's even clearer.

Stebalien (approved):

PR #376, opened 2024-04-30
Fix broken examples link in javascript.md

GGAlanSmithee:
The current link gives a 404, so I changed it to what I think is the new examples location. As someone new (returning) to libp2p, this is an important link, as it is a natural next step but not easily discovered. 

I also changed the wording slightly since it was an ill-formed sentence and indicated that the source is a part of the main libp2p repo, which is not the case any more.


GGAlanSmithee:
I see this change was also proposed in #373 but the PRs vary slightly, so I'll let maintainers decide which makes more sense

p-shahi:
closing in favor of #373

